PR #2, opened 2017-01-31
Implement pagination

aslagle:


src/ColumnsPaginator.ts
```diff
@@ -0,0 +1,93 @@
...
+        let body = this.iframe.contentDocument.body as any;
+        body.style.columnWidth = this.iframe.style.width;
+        body.style.WebkitColumnWidth = this.iframe.style.width;
+        body.style.MozColumnWidth = this.iframe.style.width;
+        body.style.height = this.iframe.style.height;
+        body.style.width = this.iframe.style.width;
+    }
+
+    public getCurrentPosition(): number {
+        let position = -(this.iframe.contentDocument.body.style.left || "0px").slice(0, -2);
```
winniequinn: Many of these variables could be declared `const` instead.

src/ColumnsPaginator.ts
```diff
@@ -0,0 +1,93 @@
...
+        let scrollWidth = this.iframe.contentDocument.body.scrollWidth;
+        let maxPosition = position + scrollWidth;
+        return position / maxPosition;
+    }
+
+    public onFirstPage(): boolean {
+        let position = -(this.iframe.contentDocument.body.style.left || "0px").slice(0, -2);
+        let width = this.iframe.contentDocument.body.offsetWidth;
```
winniequinn: Perhaps we can factor out `position` and `width` to their own functions with more descriptive names to avoid code duplication?

src/ColumnsPaginator.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    public goToNextPage(): void {
+        let position = -(this.iframe.contentDocument.body.style.left || "0px").slice(0, -2);
+        let width = this.iframe.contentDocument.body.offsetWidth;
+        let newPosition = position + width;
+        this.iframe.contentDocument.body.style.left = -newPosition + "px";
+    }
+
+    public goToPosition(position: number) {
```
winniequinn: We should document that `position` must be in range [0..1].

src/IFrameNavigator.ts
```diff
@@ -1,158 +1,388 @@
...
         let startLink = element.querySelector("a[rel=start]");
         let contentsLink = element.querySelector("a[rel=contents]");
         let navigation = element.querySelector("div[class=controls]");
+        let links = element.querySelector("ul[class=links]");
+        let linksToggle = element.querySelector("div[class=links-toggle]");
+        let previousPageLink = element.querySelector("div[class=previous-page]");
+        let nextPageLink = element.querySelector("div[class=next-page]");
+        let loading = element.querySelector("div[class=loading]");
 
-        if (!iframe || !nextLink || !previousLink || !startLink || !contentsLink || !navigation ||
```
winniequinn: It might be better to have a helper function that wraps `querySelector` which returns a non-null value or throws an exception if there's an internal inconsistency.

src/IFrameNavigator.ts
```diff
@@ -1,158 +1,388 @@
...
-            !(nextLink instanceof HTMLAnchorElement) ||
-            !(previousLink instanceof HTMLAnchorElement) ||
+        if (!iframe || !nextChapterLink || !previousChapterLink ||
+            !startLink || !contentsLink || !navigation || !links ||
+            !loading ||
+            (this.paginator && !linksToggle) ||
+            (this.paginator && !previousPageLink) ||
+            (this.paginator && !nextPageLink) ||
+            !(nextChapterLink instanceof HTMLAnchorElement) ||
```
winniequinn: We could handle these with `as` where they're first assigned.

src/IFrameNavigator.ts
```diff
@@ -1,158 +1,388 @@
...
+            } else {
+                nextElement = nextElement.parentElement;
+            }
+        }
+        if (foundLink) {
+            let link = (foundLink as HTMLAnchorElement);
+            let isSameOrigin = (
+                window.location.protocol === link.protocol &&
+                window.location.port === link.port &&
+                window.location.hostname == link.hostname
```
winniequinn: ===

src/IFrameNavigator.ts
```diff
@@ -1,158 +1,388 @@
...
+        let position = {
+            resource: this.contentsLink.href,
+            position: 0
+        };
+        this.navigate(position);
+        event.preventDefault();
+    }
+
+    private navigate(readingPosition: ReadingPosition): void {
+        this.loading.style.display = "block";
```
winniequinn: It would be nice if we displayed this after N milliseconds to avoid it flashing on every load.

winniequinn (changes requested):

aslagle:
@winniequinn I addressed all your comments

winniequinn (approved):

PR #3, opened 2017-02-01
Implemented table of contents using TOC from manifest instead of contents link

aslagle:


src/IFrameNavigator.ts
```diff
@@ -341,14 +373,30 @@ export default class IFrameNavigator implements Navigator {
     }
 
     private handleContentsClick(event: MouseEvent): void {
-        const position = {
-            resource: this.contentsLink.href,
-            position: 0
-        };
-        this.navigate(position);
+        const display: string | null = this.toc.style.display;
+        if (display && display === "none") {
```
winniequinn: We can just do `display === "none"`.

winniequinn (changes requested):
As discussed, I think it would be better if the TOC were full-screen. It would make it easier to have UI for closing the TOC, would make dealing with accessibility issues somewhat easier, and would allow for a less cramped view on mobile. Our current apps for Android and iOS are a good source of inspiration here.

src/IFrameNavigator.ts
aslagle: fixed

aslagle:
I changed the example in this repo to have a full-screen TOC, and it's scrollable now.

winniequinn (approved):

PR #4, opened 2017-02-06
Storage

aslagle:
I added a new Storage interface since there were multiple modules using localStorage and I'm about to add more. There are two implementations, one that uses memory and will lose everything if you close the page, and one that uses localStorage but falls back to the memory implementation if it's not available. This makes it possible to use in private mode in iOS safari.


src/Annotator.ts
```diff
@@ -1,5 +1,5 @@
 interface Annotator {
-    start(manifestUrl: string): Promise<void>;
+    start(): Promise<void>;
```
winniequinn, line 2: After talking to @aslagle, it seems like we can remove `start` from this interface.

src/LocalAnnotator.ts
```diff
@@ -0,0 +1,31 @@
+import Annotator from "./Annotator";
+import Store from "./Store";
+
+/** Annotator that stores annotations locally, in the browser. */
+export default class LocalAnnotator implements Annotator {
+    private store: Store;
+    private LAST_READING_POSITION = "last-reading-position";
```
winniequinn, line 7: Is there a way to declare this constant somehow?

src/LocalStorageStore.ts
```diff
@@ -0,0 +1,54 @@
+import Store from "./Store";
+import MemoryStore from "./MemoryStore";
+
+/** Class that stores key/value pairs in localStorage if possible
+    but falls back to an in-memory store. */
+export default class LocalStorageStore implements Store {
+    private enabled: boolean;
```
winniequinn: It seems like `enabled` is really `isUsingLocalStorage`. Perhaps this can be better named.

src/LocalStorageStore.ts
```diff
@@ -0,0 +1,54 @@
...
+    
+    public constructor() {}
+
+    public async start(manifestUrl: string): Promise<void> {
+        this.manifestUrl = manifestUrl;
+        try {
+            // In some browsers (eg iOS Safari in private mode), 
+            // localStorage exists but throws an exception when
+            // you try to write to it.
+            window.localStorage.setItem("test", "test");
```
winniequinn: Just to be an extra 0.001% better, maybe we can store a unique string instead (e.g. a UUID). This would prevent overwriting something that someone named `test`, of course.

src/LocalStorageStore.ts
```diff
@@ -0,0 +1,54 @@
...
+    private getLocalStorageKey(key: string): string {
+        return this.manifestUrl + "-" + key;
+    }
+
+    public async get(key: string): Promise<string | null> {
+        let value: string | null = null;
+        if (this.enabled) {
+            value = window.localStorage.getItem(this.getLocalStorageKey(key));
+        }
+        if (this.memoryStore) {
```
winniequinn: These conditionals can be cleaned up a bit when `this.enabled` is either renamed or removed (because we just will check if `this.memoryStore` is non-null).

src/ServiceWorkerCacher.ts
```diff
@@ -1,15 +1,18 @@
 import Cacher from "./Cacher";
+import Store from "./Store";
 import Manifest from "./Manifest";
 
 /** Class that caches URLs using ServiceWorker's Cache API. */
```
winniequinn: It might be better to say this caches responses, not URLs.

src/ServiceWorkerCacher.ts
```diff
@@ -1,15 +1,18 @@
...
 export default class ServiceWorkerCacher implements Cacher {
     private serviceWorkerPath: string;
+    private store: Store;
     private supported: boolean;
```
winniequinn: This could be given a more descriptive name. Perhaps `areServiceWorkersSupported` is better?

src/ServiceWorkerCacher.ts
```diff
@@ -1,15 +1,18 @@
...
 
     /** Create a ServiceWorkerCacher. */
     /** @param serviceWorkerPath Location of the service worker js file. */
-    public constructor(serviceWorkerPath: string = "sw.js") {
+    public constructor(store: Store, serviceWorkerPath: string = "sw.js") {
```
winniequinn, line 13: I am not sure if `serviceWorkerPath` should have a default argument. `"sw.js"` happens to be the correct value for us, but it's not a value that would apply to anyone else, I don't think. I am not sure if I understand this correctly though!

src/WebpubViewer.ts
```diff
@@ -1,18 +1,22 @@
...
     private navigator: Navigator;
     private annotator: Annotator | null;
 
     /** Create a WebpubViewer. */
+    /** @param store A Store to save application settings or data. */
     /** @param cacher A Cacher to handle storing webpub and app resources for offline use. */
     /** @param navigator A Navigator to handle moving between spine items and resources of the publication. */
     /** @param annotator Optional Annotator that handles bookmarking, last reading position, highlights. */
-    public constructor(cacher: Cacher, navigator: Navigator, annotator: Annotator | null = null) {
+    public constructor(store: Store, cacher: Cacher, navigator: Navigator, annotator: Annotator | null = null) {
```
winniequinn, line 18: Would making `annotator` an optional argument be slightly cleaner?

winniequinn (changes requested):
In addition to the line-by-line comments below, I think the `start` component of the various service interfaces should be removed. Instead, whatever constructs the service in question should pass in whatever is needed (e.g. a manifest URL) at the point of construction.

src/Annotator.ts
aslagle: I created an issue for doing this for all the interfaces https://github.com/NYPL-Simplified/webpub-viewer/issues/5

src/LocalAnnotator.ts
aslagle: sadly I don't think so

src/ServiceWorkerCacher.ts
aslagle: This default works if you put the service worker at the root of your exploded epub, which I think is the simplest setup. It's hard to say what would be most common.

It works for the example in this repo because I have a symlink to sw.js there.

It actually doesn't work for our real use case (https://github.com/NYPL-Simplified/web-reader-landing-page/blob/master/book_index.html.template#L108), which is why I had to make it configurable.

src/WebpubViewer.ts
aslagle: Yes, for now, but I will probably have to add another optional service soon and then I won't be able to do that.

src/LocalStorageStore.ts
aslagle: removed

src/LocalStorageStore.ts
aslagle: done

src/LocalStorageStore.ts
aslagle: removed enabled

src/ServiceWorkerCacher.ts
aslagle: done

src/ServiceWorkerCacher.ts
aslagle: done

src/ServiceWorkerCacher.ts
winniequinn: Understood! Thanks for explaining!

winniequinn (approved):

Issue #6, opened 2017-02-10
Implement reading improvements


leonardr:
The following need to be improved:

* Scrolling and pagination settings
* Settings for font size and margins
* Determine whether the book is fully downloaded and available offline, and show this information to the user.
* Add instructions about pagination controls and adding the book to your home screen.
* Redesign the navigation header

aslagle:
This is done except "Add instructions about pagination controls and adding the book to your home screen." I don't think we need pagination control instructions if we implement swiping, and I think we can hold off on the home screen thing for now since I'm still not sure if or when it would be useful.

PR #9, opened 2017-02-14
Scrolling/pagination controls

### LINKED ISSUE #5: Remove start method from interfaces
See https://github.com/NYPL-Simplified/webpub-viewer/pull/4#pullrequestreview-20342285

aslagle:
This has the first part of #6 and also fixes #5.

It adds a settings button to the controls that brings up links to toggle between scrolling and pagination.


examples/alice/webpub-viewer.js
```diff
@@ -0,0 +1 @@
+../../dist/webpub-viewer.js
```
aslagle, line 1: This was to make it easier to test the service worker.

examples/alice/index.html
```diff
@@ -53,15 +57,19 @@
...
-      var store = new LocalStorageStore.default();
-      var cacher = new ServiceWorkerCacher.default(store);
-      var paginator = new ColumnsPaginator.default();
-      var annotator = new LocalAnnotator.default(store);
-      var navigator = new IFrameNavigator.default(cacher, paginator, annotator);
-      var viewer = new WebpubViewer.default(store, cacher, navigator, annotator);
-      viewer.start(element);
+      var manifestUrl = new URL("manifest.json", window.location.href).href;
+
+      var store = new LocalStorageStore.default(manifestUrl);
```
winniequinn: It might be better if `manifestUrl` were actually of type `URL`. It would ensure that `LocalStorageCache`, `ServiceWorkerCacher`, and so on are at least initialized with well-formed URLs. They could, of course, internally call the `href` method on the URL if they only need it as a string.

examples/alice/index.html
```diff
@@ -53,15 +57,19 @@
...
+      ServiceWorkerCacher.default.create(store, manifestUrl).then(function (cacher) {
+        var paginator = new ColumnsPaginatedBookView.default();
+        var scroller = new ScrollingBookView.default();
+        var annotator = new LocalAnnotator.default(store);
+        BookSettings.default.create([paginator, scroller], store).then(function (settings) {
```
winniequinn, line 69: I really like how initialization was changed to use promises!

src/BookSettings.ts
```diff
@@ -0,0 +1,120 @@
...
+    <li><a href='#' class='${className}'>${label}</a></li>
+`;
+
+export default class BookSettings extends HTMLView {
+    private store: Store;
+    private bookViews: BookView[];
+    private viewLinks: { [key: string]: HTMLAnchorElement };
+    private selectedView: BookView;
+    private viewChangeCallback: () => void;
+    private SELECTED_VIEW_KEY = "settings-selected-view";
```
winniequinn: You can declare `SELECTED_VIEW_KEY` with the `private static readonly` qualifiers and change all references from `this.SELECTED_VIEW_KEY` to `BookSettings. SELECTED_VIEW_KEY`. This should actually give you a proper constant!

src/BookSettings.ts
```diff
@@ -0,0 +1,120 @@
...
+        ${options}
+    </ul></li>
+`;
+
+const optionTemplate = (className: string, label: string) => `
+    <li><a href='#' class='${className}'>${label}</a></li>
+`;
+
+export default class BookSettings extends HTMLView {
+    private store: Store;
```
winniequinn: Many of these properties are only set in the constructor. As such, you can declare them `private readonly` which will make this explicit. (The `readonly` qualifier restricts mutation to occurring within the constructor of the class only. It is essentially `const` for instance variables.)

src/BookSettings.ts
```diff
@@ -0,0 +1,120 @@
...
+                        break;
+                    }
+                }
+            }
+            this.selectedView = selectedView;
+        }
+    }
+
+    public renderControls(element: HTMLElement): void {
+        let sections = [];
```
winniequinn: Also picky, but you can declare `sections` as `const` here: You're mutating the array `sections` is initialized to, not the variable `sections itself.

src/BookSettings.ts
```diff
@@ -0,0 +1,120 @@
...
+
+        if (this.bookViews.length > 1) {
+            let options = [];
+            for (const bookView of this.bookViews) {
+                options.push(optionTemplate(bookView.name, bookView.label));
+            }
```
winniequinn: There's a more pleasant and functional way to write this, for what it's worth:

```typescript
const options = this.bookViews.map(bookView =>
    optionTemplate(bookView.name, bookView.label)
);
```

src/BookSettings.ts
```diff
@@ -0,0 +1,120 @@
...
+            const link = this.viewLinks[view.name];
+            if (link) {
+                link.addEventListener("click", (event: MouseEvent) => {
+                    const position = this.selectedView.getCurrentPosition();
+                    this.selectedView.stop();
+                    view.start(position);
+                    this.selectedView = view;
+                    this.updateViewLinks();
+                    this.storeSelectedView(view);
+                    if (this.viewChangeCallback()) {
```
winniequinn: I think you meant `if (this.viewChangeCallback)` here: I don't think you want to call it to see if you have it!

src/BookView.ts
```diff
@@ -0,0 +1,11 @@
+interface BookView {
+    name: string;
+    label: string;
+    setBookElement(element: Element): void;
+    setTopMargin(number: number): void;
+    start(position: number): void;
```
winniequinn: It would be good to document what `start` and `stop` are meant to do here.

src/ColumnsPaginatedBookView.ts
```diff
@@ -1,13 +1,24 @@
...
+import PaginatedBookView from "./PaginatedBookView";
 
-export default class ColumnsPaginator implements Paginator {
+export default class ColumnsPaginatedBookView implements PaginatedBookView {
     private iframe: HTMLIFrameElement;
+    private topMargin: number = 0;
 
-    public start(iframe: HTMLIFrameElement, position: number): Promise<void> {
+    public name = "columns-paginated-view"
+    public label = "Paginated View"
```
winniequinn: These should be `readonly`, I think!

src/HTMLView.ts
```diff
@@ -0,0 +1,14 @@
+export default class HTMLView {
+    protected findElement(parentElement: Element, selector: string): Element | null {
+        return parentElement.querySelector(selector);
+    }
+
+    protected findRequiredElement(parentElement: Element, selector: string): Element {
```
winniequinn: It would be good to document that this method may throw an exception.

src/HTMLView.ts
```diff
@@ -0,0 +1,14 @@
+export default class HTMLView {
```
winniequinn: I am not sure `HTMLView` should be a class. It seems like it is just a couple of utility methods.

Perhaps it would be better to change `HTMLView` to be a module with two exported functions and change `BookSettings` and `IFrameNavigator` to simply use the functions as needed? Nothing seems to depend on anything being a subclass of `HTMLView`.

If you do want it to remain a class for some reason, I would consider declaring its two methods `static`.

winniequinn (changes requested):
I noticed a weird visual bug in desktop Safari where toggling the pagination mode causes the outermost bullet to disappear. Everything else appears to be okay though!

examples/alice/index.html
aslagle: yay!

src/BookView.ts
aslagle: I tried.... it's hard to explain.

aslagle:
@winniequinn I addressed most of your comments!

I'm not too worried about the bullet disappearing since the styling will be changing significantly.

I opened a separate issue (#10) for the other bug you found and I may come back to it after I implement margins.

winniequinn (approved):

PR #13, opened 2017-02-17
Font size setting

aslagle:
This is the second part of #6.

Setting the font size automatically adjusts the side margins.

### DIFF
--- a/src/BookSettings.ts
+++ b/src/BookSettings.ts
@@ -36,11 +36,12 @@ export default class BookSettings {
 
     /** @param store Store to save the user's selections in. */
     /** @param bookViews Array of BookView options. */
-    /** @param fontSizes Array of font size strings sorted from smallest to largest. */
-    /** @param defaultFontSize Initial font size to use until the user makes a selection. */
-    public static async create(store: Store, bookViews: BookView[], fontSizes: string[], defaultFontSize?: string) {
+    /** @param fontSizesInPixels Array of font sizes in pixels sorted from smallest to largest. */
+    /** @param defaultFontSizeInPixels Initial font size to use until the user makes a selection. */
+    public static async create(store: Store, bookViews: BookView[], fontSizesInPixels: number[], defaultFontSizeInPixels?: number) {
+        const fontSizes = fontSizesInPixels.map(fontSize => fontSize + "px");
         const settings = new this(store, bookViews, fontSizes);
-        await settings.initializeSelections(defaultFontSize);
+        await settings.initializeSelections(defaultFontSizeInPixels ? defaultFontSizeInPixels + "px" : undefined);
         return settings;
     }
 
--- a/src/BookSettings.ts
+++ b/src/BookSettings.ts
@@ -34,20 +34,23 @@ export default class BookSettings {
     private static readonly SELECTED_VIEW_KEY = "settings-selected-view";
     private static readonly SELECTED_FONT_SIZE_KEY = "settings-selected-font-size";
 
+    /** @param store Store to save the user's selections in. */
+    /** @param bookViews Array of BookView options. */
     /** @param fontSizes Array of font size strings sorted from smallest to largest. */
-    public static async create(bookViews: BookView[], fontSizes: string[], store: Store) {
-        const settings = new this(bookViews, fontSizes, store);
-        await settings.initializeSelections();
+    /** @param defaultFontSize Initial font size to use until the user makes a selection. */
+    public static async create(store: Store, bookViews: BookView[], fontSizes: string[], defaultFontSize?: string) {
+        const settings = new this(store, bookViews, fontSizes);
+        await settings.initializeSelections(defaultFontSize);
         return settings;
     }
 
-    protected constructor(bookViews: BookView[], fontSizes: string[], store: Store) {
+    protected constructor(store: Store, bookViews: BookView[], fontSizes: string[]) {
+        this.store = store;
         this.bookViews = bookViews;
         this.fontSizes = fontSizes;
-        this.store = store;
     }
 
-    private async initializeSelections(): Promise<void> {
+    private async initializeSelections(defaultFontSize?: string): Promise<void> {
         if (this.bookViews.length >= 1) {
             let selectedView = this.bookViews[0];
             const selectedViewName = await this.store.get(BookSettings.SELECTED_VIEW_KEY);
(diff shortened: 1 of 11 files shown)


src/ColumnsPaginatedBookView.ts
```diff
@@ -143,7 +166,7 @@ export default class ColumnsPaginatedBookView implements PaginatedBookView {
         this.setLeftColumnsWidth(0);
 
         const width = this.getColumnWidth();
-        const rightWidth = this.getRightColumnsWidth();
+        let rightWidth = this.getRightColumnsWidth();
```
winniequinn: I think maybe this was an inadvertant change. It can be left as `const`, it seems.

src/IFrameNavigator.ts
```diff
@@ -347,11 +352,22 @@ export default class IFrameNavigator implements Navigator {
...
-        selectedView.goToPosition(oldPosition);
-    }
 
-    private async handleScroll(): Promise<void> {
-        return this.saveCurrentReadingPosition();
+        const fontSize = this.settings.getSelectedFontSize();
+        this.iframe.contentDocument.body.style.fontSize = fontSize;
+        this.iframe.contentDocument.body.style.lineHeight = "1.5";
+
+        const sideMargin = parseInt(fontSize.slice(0, -2)) * 2;
```
winniequinn, line 360: This seems to work, but it also seems that it assumes font sizes are set in terms of pixels. If this is the only place that assumption is made, it might be better to not depend on it so font sizes can also be given in terms of percentages or other metrics. I do not think the margin necessarily needs to change with the font size. (Ideally, it would be independently adjustable.)

src/ScrollingBookView.ts
```diff
@@ -11,19 +11,23 @@ export default class ScrollingBookView implements BookView {
         this.iframe = iframe;
     }
 
-    public setTopMargin(topMargin: number): void {
-        this.topMargin = topMargin;
+    public setSideMargin(margin: number) {
+        this.sideMargin = margin;
```
winniequinn: You probably know this, but there is not a need to have private instance variables with public setters like this in TypeScript. Instead of this:

```typescript
private sideMargin: number = 0;

public setSideMargin(margin: number) {
    this.sideMargin = margin;
}
```

I would just do this:

```typescript
sideMargin: number = 0;
```

If you then later decide you want to have the setter (or getter!) perform some action, you can refactor the code as such without breaking the public API:

```typescript
private _sideMargin: number = 0;

get sideMargin(): number {
     // Do whatever here!
    return this._sideMargin;
}

set sideMargin(margin: number) {
    this._sidemargin = margin;
    // Do whatever here!
}
```

The advantages of doing this way are that is avoids the extra code until you're sure you need it and it lets people use property syntax (e.g. `scrollingBookview.sideMargin = 8`) for getters and setters.

src/BookSettings.ts
```diff
@@ -51,16 +61,30 @@ export default class BookSettings {
             }
             this.selectedView = selectedView;
         }
+
+        if (this.fontSizes.length >= 1) {
+            let selectedFontSize = await this.store.get(BookSettings.SELECTED_FONT_SIZE_KEY);
+            if (!selectedFontSize) {
+                const defaultFontSizeIndex = Math.floor(this.fontSizes.length / 2);
```
winniequinn: I think it would be nice to be able to set the default font size explicitly rather than assume it's whatever the font size in the middle of `this.fontSizes` happens to be. Perhaps `create` can also take an `initialFontSize` argument and it can throw an exception (or default to a size in the middle) if the initial size is not within the array?

src/BookSettings.ts
```diff
@@ -22,18 +22,28 @@ export default class BookSettings {
...
-        const settings = new this(bookViews, store);
+    /** @param fontSizes Array of font size strings sorted from smallest to largest. */
+    public static async create(bookViews: BookView[], fontSizes: string[], store: Store) {
+        const settings = new this(bookViews, fontSizes, store);
         await settings.initializeSelections();
         return settings;
     }
 
-    protected constructor(bookViews: BookView[], store: Store) {
+    protected constructor(bookViews: BookView[], fontSizes: string[], store: Store) {
```
winniequinn: It might be a good idea to document that the order of the elements in `fontSizes` will determine the order in which they will be cycled through. It can be reasonably inferred, but still!

winniequinn (changes requested):

src/IFrameNavigator.ts
aslagle: I talked to Mauricio and Ricardo about this and they thought it was better not to have the margin be independently adjustable. We weren't sure what it should be, but this seemed okay and we can play with it later. Ideally, it would be set to 2rem and we wouldn't need the pixel value, but I need to compute the size of the iframe in pixels and I don't know how else I could do that. The font size is in pixels because all the other proportional measurements in the CSS use that base font size.

winniequinn: Got it! Sounds good to me!

src/ColumnsPaginatedBookView.ts
aslagle: It was, thanks :)

src/ScrollingBookView.ts
aslagle: I think I did it that way because I wanted to have a setter with no getter, but I like your way better. And I didn't know about TypeScript's get and set syntax.

src/BookSettings.ts
aslagle: I put the documentation of this on create instead of the constructor.

src/BookSettings.ts
aslagle: Done.

src/BookSettings.ts
```diff
@@ -34,20 +34,23 @@ export default class BookSettings {
...
     private static readonly SELECTED_FONT_SIZE_KEY = "settings-selected-font-size";
 
+    /** @param store Store to save the user's selections in. */
+    /** @param bookViews Array of BookView options. */
     /** @param fontSizes Array of font size strings sorted from smallest to largest. */
-    public static async create(bookViews: BookView[], fontSizes: string[], store: Store) {
-        const settings = new this(bookViews, fontSizes, store);
-        await settings.initializeSelections();
+    /** @param defaultFontSize Initial font size to use until the user makes a selection. */
+    public static async create(store: Store, bookViews: BookView[], fontSizes: string[], defaultFontSize?: string) {
```
winniequinn: It might be better to just take a `fontSizesinPixels: [number]` and do `fontSizesInPixels.map(n => n + "px")` internally so people don't accidentally pass font sizes like "100%" or "14pt".

winniequinn (changes requested):

src/BookSettings.ts
aslagle: Done

winniequinn (approved):

PR #15, opened 2017-02-27
Sass too

ricardoom:
@aslagle I set up sass, check it out.


examples/alice/css/core.css
```diff
@@ -1,75 +1,69 @@
-@charset "UTF-8";
```
aslagle, line 1: Just fyi, the files in examples/alice/css are extracted from an epub file, and we won't have control over them when we use this with other books. I'd probably leave them alone but it doesn't really matter.

ricardoom: so any global stylings need to happen in the root then, yeah?

aslagle: Any styling changes _within the book content_ have to be done in js, the way I have it set up now. For the nav and any of the book controls, I think we should include a new css file from index.html.

ricardoom:
@aslagle could you take a look at this? Do you think it can be merged? I could branch off of this and work on the node-sass bit. not sure about the sequencing...

aslagle:
I would like to get the node-sass commands in first, if you don't mind, to make sure I can build it.

I also don't think I want the NYPL reds in here since this is a general library. If we do an NYPL-branded version I'd make another repo that uses this library but has its own colors. The NYPL Monochrome is good though.

src/styles/css/controls.css
```diff
@@ -0,0 +1,65 @@
...
+  width: 12.5rem;
+  height: 12.5rem;
+  text-align: center;
+  left: calc(50% - 6.25rem);
+  top: calc(50% - 6.25rem);
+  background-color: #76777b; }
+
+.controls-view {
+  position: fixed;
+  top: 0.125rem;
```
aslagle: This is making the settings view cover the nav, so you can't click it again to close the settings, which right now is the only way to close the settings.

ricardoom: ok. i fixed the issue where the settings dialog was hiding the settings link.

PR #16, opened 2017-02-28
Offline settings option

aslagle:
This branch adds an option in the settings to enable downloading the book for offline use. In addition, I think we'll want to have a popup that asks the user if they want to download when they first open the book, but that will be a separate branch.

I also improved the way the app cache fallback works. If service workers are available, it won't use the app cache at all. If they're not, it loads an iframe that starts the app cache.

Once the user has enabled offline use, the settings shows the download status but doesn't provide any options to change anything.

This is the third part of #6.


aslagle:
I went ahead and added the popup when you first open a book. I think we'll eventually want something nicer than the native confirm dialog, but we can start getting feedback on this.

src/BookSettings.ts
```diff
@@ -18,21 +18,33 @@ const optionTemplate = (className: string, label: string) => `
...
+    private offlineLink: HTMLAnchorElement;
+    private offlineStatusElement: HTMLElement;
 
     private viewChangeCallback: () => void;
     private fontSizeChangeCallback: () => void;
+    private offlineEnabledCallback: () => void;
 
     private selectedView: BookView;
     private selectedFontSize: string;
+    private offlineEnabled: boolean | null;
```
winniequinn: This could be more self-documenting if we used an enum instead, e.g.:

```typescript
enum OfflineStatus = {
  Enabled,
  Disabled,
  Unselected
}

private offlineStatus: OfflineStatus;
```

src/BookSettings.ts
```diff
@@ -210,11 +257,28 @@ export default class BookSettings {
         return this.selectedFontSize;
     }
 
+    public getOfflineEnabled(): boolean {
+        if (this.offlineEnabled === null) {
+            const enable = window.confirm("Would you like to download this book to read offline?");
```
winniequinn: I wouldn't expect this method to block on UI. Perhaps a better name for the method would be sufficient?

src/Cacher.ts
```diff
@@ -1,7 +1,12 @@
 import Manifest from "./Manifest";
 
 interface Cacher {
-    getManifest(manifestUrl: URL): Promise<Manifest>;
+    getManifest(manifestUrl: URL): Promise<Manifest>;
```
winniequinn: After discussing this with @aslagle, it seems like `getManifest` need not be part of `Cacher` functionality and can be moved out to another location.

I should also note that the `getManifest` method of `Cacher` does _not_ return a cache manifest as I first thought! It instead returns the manifest for the book that contains spine items and the like. Perhaps renaming this method to `getJSONManifest` or similar would make things more clear.

src/Cacher.ts
```diff
@@ -1,7 +1,12 @@
 import Manifest from "./Manifest";
 
 interface Cacher {
-    getManifest(manifestUrl: URL): Promise<Manifest>;
+    getManifest(manifestUrl: URL): Promise<Manifest>;   
+
+    // This should kick off the downloads to enable offline use.
+    enable(): Promise<void>;
+
+    renderStatus(element: HTMLElement): void;
```
winniequinn: I think it might be better if `Cacher` is not responsible for view-level code. Instead, I might have have something like this:

```typescript
enum RenderStatus {
    Idle,
    Downloading,
    Downloaded
};

onRenderStatusChange(callback: (status: RenderStatus) => void);
```

If you wanted to communicate some kind of progress percentage (which may not be possible), you could use a discriminated union approach instead:

```typescript
interface Idle {
    kind: "Idle";
}
interface Downloading {
    kind: "Downloading";
    progress: number; // [0, 1]
}

interface Downloaded {
    kind: "Downloaded";
}

type RenderStatus
    = Idle
    | Downloading
    | Downloaded;

onRenderStatusChange(callback: (status: RenderStatus) => void);
```

The callback could then match on `status.kind` and have type-safe access to `status.progress` if and only if `kind === "Downloading"`.

src/ServiceWorkerCacher.ts
```diff
@@ -1,36 +1,54 @@
...
+    falls back to the application cache if service workers aren't available. */
 export default class ServiceWorkerCacher implements Cacher {
     private readonly serviceWorkerPath: string;
     private readonly store: Store;
+    private readonly manifestUrl: URL;
     private readonly areServiceWorkersSupported: boolean;
+    private readonly fallbackCacher: ApplicationCacheCacher | null;
+    private statusElement: HTMLDivElement;
+    private cachingStarted: boolean = false;
+    private cachingFinished: boolean = false;
```
winniequinn: `cachingStarted` and `cachingFinished` can probably be collapsed into a single member of an enum type (e.g. `NotYetStarted`, `Started`, and `Finished`) so we don't have to worry about maintaining some kind of `cachingFinished == true` ==> `cachingStarted == true` invariant.

winniequinn (changes requested):

aslagle:
@winniequinn thanks for the comments :) 

This is ready for you to take another look!

src/BookSettings.ts
```diff
@@ -25,6 +25,12 @@ const offlineTemplate = `
     </li>
 `;
 
+enum OfflineStatus {
+    ENABLED,
```
winniequinn: I think it is more common to use PascalCase for enums. It seems to be recommended by the [official coding guidelines](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines) as well.

src/Cacher.ts
```diff
@@ -1,8 +1,30 @@
+export enum CacheStatus {
+    // The book has not been cached.
```
winniequinn: I might prefer JSDoc comments for these sort of things (`/** … */`) so they show up in generated documentation and in people's IDEs. I love this change otherwise though!

src/IFrameNavigator.ts
```diff
@@ -114,7 +115,7 @@ export default class IFrameNavigator implements Navigator {
             this.settings.onViewChange(this.updateBookView.bind(this));
             this.settings.onFontSizeChange(this.updateFontSize.bind(this));
             this.settings.onOfflineEnabled(this.enableOffline.bind(this));
-            this.cacher.renderStatus(this.settings.getOfflineStatusElement());
+            this.cacher.onStatusUpdate(this.updateOfflineCacheStatus.bind(this));
```
winniequinn: This approach makes me happy. ^_^

winniequinn (approved):
Aside from a couple of minor things I commented upon that you may or may not want to change, I think this is ready to go!

PR #20, opened 2017-03-20
Nav improve

ricardoom:
Wrapping up the stylings here. After review and comments on this branch. If there is nothing major that needs doing, I think it should be merged and any further design work can take place on other branches.


aslagle:
I need to update the tests for a couple of the changes I made on this branch before it's merged.

mgiraldo:
screenshots?

aslagle:
I updated the demo http://dev-mta.librarysimplified.org/examples/alice/

mgiraldo:
maybe have a sample “Aa” in the settings so user knows how much bigger/smaller the font is? Right now the settings cover a good portion of the book... or put a semi-transparent background to the settings?

src/styles/sass/main.scss
```diff
@@ -6,87 +6,319 @@
...
-  background-color: $nypl-gray-cool;
+  width: 12.5rem;
 }
 
 .controls-view {
+  font-family: $sans-serif-stack;
   height: calc(100% - 2rem);
   left: 0;
   overflow: scroll;
   position: fixed;
```
mgiraldo: add this rule: 

`-webkit-overflow-scrolling:touch`

so that iOS shows the smooth-enabled scrolling… this is iOS only but still we should take advantage of it

ricardoom:
page settings l
page l
page settings m
page sm
page settings sm
contents-m
contents-sm

src/BookSettings.ts
```diff
@@ -3,19 +3,19 @@ import * as HTMLUtilities from "./HTMLUtilities";
...
 const sectionTemplate = (sectionName: string, options: string) => `
-    <li>${sectionName}<ul>
+    <li>${sectionName}<ul class="settings-options">
         ${options}
     </ul></li>
 `;
 
 const optionTemplate = (className: string, label: string) => `
-    <li><a href='#' class='${className}'>${label}</a></li>
+    <li><a href='#' class='${className}' role="button">${label}</a></li>
```
aslagle: should this be an actual button?

src/IFrameNavigator.ts
```diff
@@ -13,19 +13,70 @@ import * as HTMLUtilities from "./HTMLUtilities";
...
+            <title id="home-button">Home Button</title>
+            <g transform="translate(0,-952.36218)">
+              <path style="text-indent:0;text-transform:none;direction:ltr;block-progression:tb;baseline-shift:baseline;color:#000000;enable-background:accumulate;" d="M 48.18752,956.9572 8.1875204,986.95679 c -0.7389,0.5602 -1.1923,1.4789 -1.1875,2.4062 l 0,55.99921 c 2e-4,1.5708 1.4292,2.9998 2.9999996,3 l 27,0 c 1.5708,-2e-4 2.9998,-1.4292 3,-3 l 0,-34.9995 20,0 0,34.9995 c 2e-4,1.5708 1.4292,2.9998 3,3 l 26.999997,0 c 1.5708,-2e-4 2.9998,-1.4292 3,-3 l 0,-55.99921 c 0,-0.9273 -0.4486,-1.846 -1.1875,-2.4062 L 51.81252,956.9572 c -1.3099,-0.8318 -2.4519,-0.75401 -3.625,0 z m 1.8125,6.15616 36.999997,27.74963 0,51.49931 -20.999997,0 0,-34.9996 c -2e-4,-1.5707 -1.4292,-2.9998 -3,-2.9999 l -26,0 c -1.5708,10e-5 -2.9998,1.4292 -3,2.9999 l 0,34.9996 -21,0 0,-51.49931 z"/>
+            </g>
+          </svg>
+            <span class="setting-text home">Start</span>
+          </a>
+        </li>
+        <li>
+          <a rel="contents" class="disabled" role="button" aria-labelledby="contents" aria-haspopup="true" aria-expanded="false">
```
aslagle: is aria-expanded supposed to change when the popup is showing?

src/styles/sass/img/contents.svg
```diff
@@ -0,0 +1,7 @@
+<svg viewBox="0 0 24 24">
```
aslagle: why are there svg files as well as the stuff in IFrameNavigator? are they both used?

src/BookSettings.ts
mgiraldo: good idea. yes

src/IFrameNavigator.ts
mgiraldo: yes. should become `true`. [this example](https://nypl.github.io/design-toolkit/sections/buttons.html#menuButton) shows a similar context (with jQuery code)

src/styles/sass/img/contents.svg
ricardoom: Ah, no, those are for reference only, they are not used. I put them there for ease of transport from one computer to the next. We can remove them from the project for tidiness.

aslagle:
In the toc, it would be nice if you could click anywhere in a line, and there could be more of a hover state. The toc link with "active" class should also be highlighted somehow - that's the chapter you're currently on.

ricardoom:
@aslagle as per your request I added hover states for the TOC. Could you add an `.active` class to the current chapter? Seems like a job for JS...

aslagle:
The active class should be there already, unless something broke it...

src/IFrameNavigator.ts
ricardoom: We should probably change these to `<button>`s as well

src/IFrameNavigator.ts
```diff
@@ -13,19 +13,70 @@ import * as HTMLUtilities from "./HTMLUtilities";
...
+            <rect x="4" y="8" width="12" height="2"/>
+            <rect x="4" y="12" width="12" height="2"/>
+            <rect x="4" y="16" width="12" height="2"/>
+            <rect x="4" y="20" width="6" height="2"/>
+            </svg>
+            <span class="setting-text contents">Contents</span>
+          </a>
+        </li>
+        <li>
+          <a class="settings" role="button" aria-labelledby="settings-menu" aria-expanded="false" aria-haspopup="true">
```
ricardoom: `<button>`

ricardoom:
ready for another review. I can post screen shots if need be.
@aslagle can you take a look when you have a moment?

src/IFrameNavigator.ts
```diff
@@ -177,13 +234,13 @@ export default class IFrameNavigator implements Navigator {
 
         let statusMessage = "";
         if (status === CacheStatus.Uncached) {
-            statusMessage = "Not available offline";
+            statusMessage = "Download";
```
aslagle: This seems weird to me, next to the read offline button. Maybe there shouldn't be an additional message in this case.

src/styles/sass/img/contents.svg
aslagle: Yeah, let's remove them, unless it's still useful for you to have them here.

src/styles/sass/main.scss
```diff
@@ -6,87 +6,340 @@
...
-.controls {
-  color: $nypl-white;
+// Temp grayscaling for ease of styling...
+body {
+  // filter: grayscale(1);
+  // filter: grayscale(0);
   font-family: $sans-serif-stack;
   font-size: $basefont;
-  height: 4.65rem;
+  margin: 0 auto !important;
```
aslagle: important?

aslagle: if you're overriding https://github.com/NYPL-Simplified/webpub-viewer/blob/master/examples/alice/index.html#L9 just change that line

ricardoom: got it.

ricardoom: just to be clear you're suggesting:
removing  the inline `style` on the `<body>` tag and change the scss to this:
```body {
  font-family: $sans-serif-stack;
  font-size: $basefont;
  margin: 0 auto;
  overflow: hidden;

  @media (min-width: 45rem) {
    width: 40rem;
  }
}

Comment (aslagle) [reply src/styles/sass/main.scss]: 👍

**Comment by aslagle** [comment]
I found a bug in my margin code - when you go to the next chapter by turning a page the top and bottom margins go away.

File: src/IFrameNavigator.ts (hunk 6 above)
Comment (ricardoom) [reply src/IFrameNavigator.ts]: I would like to test removing all the labels and see if users have trouble with the controls...  what do you think about that?

Comment (aslagle) [reply src/IFrameNavigator.ts]: We can try it, but I think it's good for a user to know whether the book is ready to use offline, vs still downloading or there was an error.

Comment (aslagle) [reply src/IFrameNavigator.ts]: I guess the status message could replace the button, and not be there when the button is showing. I don't think we need labels for the view and font size controls.

Comment (ricardoom) [reply src/IFrameNavigator.ts]: I've removed the `View` & `Font Size` in my latest commit.

File: src/BookSettings.ts [hunk 8]
```diff
@@ -116,12 +116,12 @@ export default class BookSettings {
             const viewOptions = this.bookViews.map(bookView =>
                 optionTemplate("", bookView.name, bookView.label)
             );
-            sections.push(sectionTemplate("View", viewOptions.join("")));
+            sections.push(sectionTemplate("", viewOptions.join("")));
```
ricardoom: @aslagle maybe you dont want an empty string here... the compiler was complaining otherwise and I didnt want to muck about too much w/ template logic, so I did it this way to get the job done...

src/BookSettings.ts
```diff
@@ -116,12 +116,12 @@ export default class BookSettings {
...
         }
 
         if (this.fontSizes.length > 1) {
             const fontSizeOptions = optionTemplate("font-setting", "decrease", "A") + optionTemplate("font-setting", "increase", "A");
-            sections.push(sectionTemplate("Font Size", fontSizeOptions));
+            sections.push(sectionTemplate("", fontSizeOptions));
```
ricardoom: same deal here...

src/BookSettings.ts
aslagle: If none of them have labels I can get rid of that argument.

ricardoom:
merge?

aslagle (approved):

PR #34, opened 2017-04-07
Within book link handling

### LINKED ISSUE #21: In-book links should go to the right page/scroll position
Standard Ebooks' Twenty Thousand Leagues is good to test with.

aslagle:
This fixes #21.

When you click a link that goes to an element in one of the chapters of the book, it sends you to a page or scroll position so that the element is visible.

The way I did this is a little messy - when a link goes to a new chapter, the iframe loads what it thinks the position should be, then we remove that and send you to the beginning of the chapter, then we send you to the element. It makes the page jump around for a second, but you end up in the right place. I'm going to create a couple more issues and come back to this later. Not many of our books have footnotes/endnotes so it's not a huge problem yet.

While I was doing this I noticed that some of the css for large screens was breaking my logic for figuring out when you clicked a link, so I had to change that so it sets a larger margin in js instead of setting the body width in css.

Here's a demo book with endnotes: http://dev-mta.librarysimplified.org/examples/twenty/
(The TOC is a little messed up, but that's an issue with the code that sets up the book, not the reader)

### DIFF
--- a/src/BookView.ts
+++ b/src/BookView.ts
@@ -14,5 +14,6 @@ interface BookView {
 
     getCurrentPosition(): number;
     goToPosition(position: number): void;
+    goToElement(elementId: string): void;
 }
 export default BookView;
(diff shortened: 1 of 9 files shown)


aslagle:
I also noticed the links are small and hard to click on android, but they do work if you manage to click them. That's a problem with the epubs.

Issue #40, opened 2017-04-14
External links should open in new tab


aslagle:
This currently works, but I'm removing that code as part of #22 and #23, and there will probably need to be additional js in the iframe to handle it.

PR #41, opened 2017-04-14
Gesture handling in iframe

### LINKED ISSUE #10: Double scrollbar after shrinking window in desktop safari
In desktop safari, when the window gets smaller, the new iframe height is set before the browser is finished reflowing the iframe content, and the height can be too small. Then the iframe body and the document body are both scrollable, and you can sometimes see two scrollbars on top of each other.

### LINKED ISSUE #22: Implement swiping to turn pages

### LINKED ISSUE #23: Tapping should show/hide nav in scrolling view

### LINKED ISSUE #36: Selecting text should work in paginated view

aslagle:
This fixes #23
This fixes #22 
This fixes #36
And I think this also fixes #10, since I changed the scrolling view so that the iframe takes the full width but its body has margins.
However, it introduces #40 which I will have to fix in another pr.

This branch adds support for swiping to turn the page on devices that support touch. I tried to avoid conflicting with any native browser gesture handling, and it mostly seems to work. On iOS, Safari does have a swipe from the very left or right edge that goes forward or back in the browser, and there doesn't seem to be anything I can do about that.

On desktop, swiping does nothing and you have to tap to turn the page, but I added a class to the iframe when you are hovering over the left or right side, and css to make that side a little darker.

As part of this, I got rid of the elements that were sitting on top of the iframe and capturing clicks, and put all the event handling into the iframe itself.

### DIFF
--- a/src/EventHandler.ts
+++ b/src/EventHandler.ts
@@ -217,15 +217,12 @@ export default class EventHandler {
     }
 
     private checkForLink = (event: MouseEvent | TouchEvent): boolean => {
-        const target = event.target;
-        if (target instanceof Element) {
-            let nextElement = event.target as Element;
-            while (nextElement && nextElement.tagName.toLowerCase() !== "body") {
-                if (nextElement.tagName.toLowerCase() === "a") {
-                    return true;
-                } else {
-                    nextElement = nextElement.parentElement;
-                }
+        let nextElement = event.target as Element;
+        while (nextElement && nextElement.tagName.toLowerCase() !== "body") {
+            if (nextElement.tagName.toLowerCase() === "a") {
+                return true;
+            } else {
+                nextElement = nextElement.parentElement;
             }
         }
         return false;
--- a/src/EventHandler.ts
+++ b/src/EventHandler.ts
@@ -1,8 +1,8 @@
 export default class EventHandler {
-    private pendingMouseEventStart: MouseEvent | null;
-    private pendingMouseEventEnd: MouseEvent | null;
-    private pendingTouchEventStart: TouchEvent | null;
-    private pendingTouchEventEnd: TouchEvent | null;
+    private pendingMouseEventStart: MouseEvent | null = null;
+    private pendingMouseEventEnd: MouseEvent | null = null;
+    private pendingTouchEventStart: TouchEvent | null = null;
+    private pendingTouchEventEnd: TouchEvent | null = null;
 
     public onLeftTap: (event: UIEvent) => void = () => {};
     public onMiddleTap: (event: UIEvent) => void = () => {};
@@ -14,12 +14,12 @@ export default class EventHandler {
     public onRightHover: () => void = () => {};
     public onRemoveHover: () => void = () => {};
 
-    public constructor() {
-        this.pendingMouseEventStart = null;
-        this.pendingMouseEventEnd = null;
-        this.pendingTouchEventStart = null;
-        this.pendingTouchEventEnd = null;
-    }
+    private static readonly CLICK_PIXEL_TOLERANCE = 10;
+    private static readonly TAP_PIXEL_TOLERANCE = 10;
+    private static readonly DOUBLE_CLICK_MS = 200;
+    private static readonly LONG_PRESS_MS = 500;
+    private static readonly DOUBLE_TAP_MS = 200;
+    private static readonly SLOW_SWIPE_MS = 500;
 
     public setupEvents(element: HTMLElement | Document) {
         if (this.isTouchDevice()) {
@@ -57,14 +57,13 @@ export default class EventHandler {
             return;
         }
 
-        const width = window.innerWidth;
-        const height = window.innerHeight;
+        const devicePixelRatio = window.devicePixelRatio;
 
-        const relativeXDistance = (this.pendingMouseEventStart.clientX - event.clientX) / width;
-        const relativeYDistance = (this.pendingMouseEventStart.clientY - event.clientY) / height;
+        const xDevicePixels = (this.pendingMouseEventStart.clientX - event.clientX) / devicePixelRatio;
+        const yDevicePixels = (this.pendingMouseEventStart.clientY - event.clientY) / devicePixelRatio;
 
         // Is the end event in the same place as the start event?
-        if (Math.abs(relativeXDistance) < 0.1 && Math.abs(relativeYDistance) < 0.1) {
+        if (Math.abs(xDevicePixels) < EventHandler.CLICK_PIXEL_TOLERANCE && Math.abs(yDevicePixels) < EventHandler.CLICK_PIXEL_TOLERANCE) {
             if (this.pendingMouseEventEnd) {
                 // This was a double click. Let the browser handle it.
                 this.pendingMouseEventStart = null;
@@ -75,7 +74,7 @@ export default class EventHandler {
             // This was a single click.
             this.pendingMouseEventStart = null;
             this.pendingMouseEventEnd = event;
-            setTimeout(this.handleClick, 200);
+            setTimeout(this.handleClick, EventHandler.DOUBLE_CLICK_MS);
             return;
         }
 
@@ -97,9 +96,6 @@ export default class EventHandler {
             return;
         }
 
-        const width = window.innerWidth;
-        const height = window.innerHeight;
-
         const startTouch = this.pendingTouchEventStart.changedTouches[0];
         const endTouch = event.changedTouches[0];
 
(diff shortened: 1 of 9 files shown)


src/EventHandler.ts
```diff
@@ -0,0 +1,248 @@
...
+    public onMiddleTap: (event: UIEvent) => void = () => {};
+    public onRightTap: (event: UIEvent) => void = () => {};
+    public onBackwardSwipe: (event: UIEvent) => void = () => {};
+    public onForwardSwipe: (event: UIEvent) => void = () => {};
+
+    public onLeftHover: () => void = () => {};
+    public onRightHover: () => void = () => {};
+    public onRemoveHover: () => void = () => {};
+
+    public constructor() {
```
winniequinn: It seems as though you could initialize these instance variables at the point of declaration and remove the constructor.

src/EventHandler.ts
```diff
@@ -0,0 +1,248 @@
...
+            // This is a multi-touch event. Ignore.
+            return;
+        }
+
+        this.pendingTouchEventStart = event;
+    }
+
+    private handleMouseEventEnd = (event: MouseEvent): void => {
+        if (!this.pendingMouseEventStart) {
+            // Somehow we got an end event without a start event. Ignore it.
```
winniequinn, line 61: Would it make sense to `this.pendingMouseEventStart = null;` here? It probably makes little difference but it would be a bit tidier, I think.

src/EventHandler.ts
```diff
@@ -0,0 +1,248 @@
...
+        }
+
+        const width = window.innerWidth;
+        const height = window.innerHeight;
+
+        const relativeXDistance = (this.pendingMouseEventStart.clientX - event.clientX) / width;
+        const relativeYDistance = (this.pendingMouseEventStart.clientY - event.clientY) / height;
+
+        // Is the end event in the same place as the start event?
+        if (Math.abs(relativeXDistance) < 0.1 && Math.abs(relativeYDistance) < 0.1) {
```
winniequinn: _(This is a very minor point!)_

On a large desktop display, ±10% is quite a large distance for it to be considered the same place. Consider that the movements of an input device like a trackpad or mouse become smaller relative to the total display as display sizes increase. Mice and trackpads are also very precise so you really only need to have a handful of pixels worth of tolerance to begin with.

Maybe settling on a sane number of pixels and making use of [window.devicePixelRatio](http://caniuse.com/#feat=devicepixelratio) would be a way to have a more consistent result across platforms regardless of screen size and pixel density? It would let you specify a rough absolute distance rather than a distance relative to the screen size.

Just a thought!

src/EventHandler.ts
```diff
@@ -0,0 +1,248 @@
...
+                // This was a double click. Let the browser handle it.
+                this.pendingMouseEventStart = null;
+                this.pendingMouseEventEnd = null;
+                return;
+            }
+
+            // This was a single click.
+            this.pendingMouseEventStart = null;
+            this.pendingMouseEventEnd = event;
+            setTimeout(this.handleClick, 200);
```
winniequinn: Is this the time in milliseconds that is used for double-click detection? If so, I think it may be  wise to increase it a bit. The default time on macOS is probably around 500 ms and I've noticed that the tolerance of the reader is far less than the system-wide setting on my machine. As such, sometimes the navigation toggles even when I have successfully double-clicked a word to highlight it.

src/EventHandler.ts
```diff
@@ -0,0 +1,248 @@
...
+        if (!startTouch) {
+            // Somehow we saved a touch event with no touches.
+            return;
+        }
+
+        const relativeXDistance = (startTouch.clientX - endTouch.clientX) / width;
+        const relativeYDistance = (startTouch.clientY - endTouch.clientY) / height;
+
+        // Is the end event in the same place as the start event?
+        if (Math.abs(relativeXDistance) < 0.1 && Math.abs(relativeYDistance) < 0.1) {
```
winniequinn: If this logic is going to be the same as for mice, it probably would be good to consolidate it a bit.

src/EventHandler.ts
```diff
@@ -0,0 +1,248 @@
...
+                // This was a long press. Let the browser handle it.
+                this.pendingTouchEventStart = null;
+                this.pendingTouchEventEnd = null;
+                return;
+            }
+
+            // This was a single tap.
+            this.pendingTouchEventStart = null;
+            this.pendingTouchEventEnd = event;
+            setTimeout(this.handleTap, 200);
```
winniequinn: For constants like this one (e.g. 200 for the double-tap threshold, 500 for the long-press threshold, et cetera), I would declare them as static class constants, preferably with a name that indicates the unit of measure (e.g. `longPressTresholdMilliseconds`). That would allow them to be tuned more easily later from a single location (if you adjust your tests as well) and would also eliminate the use of magic numbers.

src/EventHandler.ts
```diff
@@ -0,0 +1,248 @@
...
+            this.onRightTap(this.pendingTouchEventEnd);
+        } else {
+            this.onMiddleTap(this.pendingTouchEventEnd);
+        }
+        this.pendingTouchEventEnd = null;
+        return;
+    }
+
+    private checkForLink = (event: MouseEvent | TouchEvent): boolean => {
+        let nextElement = event.target as Element;
```
winniequinn: Since `event.target` can also be a document or window, at least in general, it would be better to write the code like this:

``` typescript
let target = event.target
if(target instanceof Element) {
  // In this block, TypeScript will know `target` has type `Element` and no
  // unsafe cast is needed.
  ...
} else {
  return false;
}
```

If you haven't seen this construct before, you may want to look at [this article](https://basarat.gitbooks.io/typescript/docs/types/typeGuard.html).

src/IFrameNavigator.ts
```diff
@@ -107,6 +103,7 @@ export default class IFrameNavigator implements Navigator {
     private scroller: ScrollingBookView | null;
     private annotator: Annotator | null;
     private settings: BookSettings;
+    private eventHandler: EventHandler | null;
```
winniequinn: This is declared nullable but it looks like it's always initialized to at least `new EventHandler()`, it seems. Perhaps the `null` case can be removed here.

src/EventHandler.ts
```diff
@@ -0,0 +1,248 @@
...
+        while (nextElement && nextElement.tagName.toLowerCase() !== "body") {
+            if (nextElement.tagName.toLowerCase() === "a") {
```
winniequinn: 💞

src/IFrameNavigator.ts
```diff
@@ -469,18 +453,15 @@ export default class IFrameNavigator implements Navigator {
...
-        event.preventDefault();
+    private handleToggleLinksClick(event: MouseEvent | TouchEvent): void {
         this.hideTOC();
-        if (!this.checkForIFrameLink(event)) {
-           this.toggleDisplay(this.links);
-           this.toggleDisplay(this.linksBottom);
-        }
+        this.toggleDisplay(this.links);
+        this.toggleDisplay(this.linksBottom);
+        event.preventDefault();
```
winniequinn, line 448: For this `preventDefault` call and the two that follow, should you also be calling `stopPropagation` here? It might not strictly make any difference now but it may avoid subtle issues later.

winniequinn (changes requested):
I love this approach! So happy! ^_^

src/EventHandler.ts
winniequinn: The above comment refers to the variables that follow! Sorry about that.

src/EventHandler.ts
aslagle: Yes, this is for double-click detection. I initially tried 500ms but it felt too slow to me, particularly for turning pages :(

src/EventHandler.ts
aslagle: I don't think it's possible to get here without it already being null?

src/EventHandler.ts
aslagle: They are similar, but I think there are enough differences that it's not worth it at this point.

src/IFrameNavigator.ts
aslagle: That's true. I think I planned to make it nullable and didn't. That wouldn't make sense with a paginator, but if you used it with only a scrolling view it would be ok. I can come back to that later if it's ever useful.

src/EventHandler.ts
aslagle: Unfortunately, this doesn't work in a real browser - `target instanceof Element` is always false.

aslagle:
Added a few more bug fixes in here - #40, #38, and #43.

src/EventHandler.ts
winniequinn: We discussed this a bit over Slack, but I think the ideal solution might be a different approach entirely:

* On non-touch platforms, always show the navigation, do no special double-click handling (so users can use it to freely select text as usual), and have no delay for clicking on the left and right sides to turn pages.
* On touch platforms, double tap should probably just be disabled. No special handling of double taps should be present otherwise and no delays should be present.

I know this is a different approach from what you've already implemented. If you think it makes sense, I think it's fine to merge what you what you have and make a new issue for implementing this instead.

src/EventHandler.ts
winniequinn: I must've had something scrambled in my brain when I wrote that comment. Please ignore. ^_^

winniequinn:
I think this is okay to merge!

aslagle:
@ricardoom did you have any thoughts on the left/right hover styling before I merge this?

ricardoom:
@aslagle I was mucking about w/ it a bit friday PM. I think it can be merged. A couple things I did notice. when you increase type size, the overlays and text overlap. Also I'd probably go for a more gentle gradient. But again, I dont think these should prevent the merge...

aslagle:
@ricardoom The overlays overlap the text because they cover the entire area where you can click to turn the page. I was wondering if it should be a smaller overlay on the edge instead.

ricardoom:
cool. like I said, its not a terrible UX, so we can go with it. The other part about making the gradient bit smoother we can do later... ¿que no?

aslagle:
Okay I'll guess I'll go ahead and merge.

PR #53, opened 2017-04-24
Made back button configurable

### LINKED ISSUE #42: Add back button option

aslagle:
This fixes #42

This adds arguments to IFrameNavigator to control the href and text of the up/back link.

By default, it won't show up, and you can pass in the arguments if you want it to be there.

We'll need to modify https://github.com/NYPL-Simplified/web-reader-landing-page/blob/master/book_index.html.template to use this, once we know what it should say and where it should go.

I don't like these arguments but I don't have a better idea right now.

### DIFF
--- a/src/IFrameNavigator.ts
+++ b/src/IFrameNavigator.ts
@@ -11,19 +11,20 @@ import { OfflineStatus } from "./BookSettings";
 import EventHandler from "./EventHandler";
 import * as HTMLUtilities from "./HTMLUtilities";
 
+const upLinkTemplate = (href: string, label: string) => `
+  <a rel="up" href='${href}' aria-label="${label}">
+    <svg width="16" height="25" viewBox="0 0 16 25" aria-labelledby="back-to-book" preserveAspectRatio="xMidYMid" role="img" class="icon">
+      <title id="up-label">${label}</title>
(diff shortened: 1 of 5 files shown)


ricardoom:
looks good to me.

ricardoom (approved):

src/IFrameNavigator.ts
```diff
@@ -102,10 +103,11 @@ export default class IFrameNavigator implements Navigator {
     private annotator: Annotator | null;
     private settings: BookSettings;
     private eventHandler: EventHandler;
+    private upUrl: URL | null;
+    private upLabel: string | null;
```
winniequinn, line 107: Since you need either both of these or neither of these (I assume), I might do the following instead—it would help you eliminate nonsensical states:

```typescript
private up: { url: URL, label: string} | null;
```

If you do this, you'll need to change the constructor too, of course.

winniequinn (changes requested):

src/IFrameNavigator.ts
aslagle: You could have an upUrl with no upLabel, but not an upLabel with no upUrl.

aslagle: I think I might come back to this later and make this constructor take one config object as its argument.

PR #62, opened 2017-04-27
Show hide with css

ricardoom:
got this to work. If you approve, i'll remove the extra markup in the IFramenavigator.ts file after we merge it into `mta-style`


ricardoom:
looks like travis is complaining... is this due to removing the inline showing / hiding functionality?

aslagle:
Yeah I will try to fix that quick

aslagle (approved):

PR #66, opened 2017-04-28
RG offline improvements

ricardoom:
dig. will need to do more work for accessibility on this me thinks....


aslagle (approved):
Looks good.

PR #71, opened 2017-04-30
Bug fixes

### LINKED ISSUE #25: Handle screen rotation on mobile

### LINKED ISSUE #67: Loading message never goes away if you change chapters when you're on a network with no internet

### LINKED ISSUE #68: In TOC, clicking current chapter should close TOC without reloading iframe

### LINKED ISSUE #70: CSS column logic doesn't work in Firefox

aslagle:
Fixes #70 
Fixes https://github.com/NYPL-Simplified/web-reader-landing-page/issues/93
Fixes #68 
Fixes #67 
Fixes https://github.com/NYPL-Simplified/web-reader-landing-page/issues/94
Fixes #25


src/ColumnsPaginatedBookView.ts
```diff
@@ -19,7 +20,7 @@ export default class ColumnsPaginatedBookView implements PaginatedBookView {
 
         // any is necessary because CSSStyleDeclaration type does not include
         // all the vendor-prefixed attributes.
-        const body = this.bookElement.contentDocument.body as any;
+        const body = HTMLUtilities.findRequiredElement(this.bookElement.contentDocument, "body") as any;
```
winniequinn, line 23: At a minimum, I would cast this to `HTMLElement` instead of `any` so the following usage is safer.

Better still though would be to change `findRequiredElement` so it throws an exception if the element it finds is not a subtype of `HTMLElement`.  (It is a member of `HTMLUtilities`, after all.) You could then give it a return type of `HTMLElement` instead of merely `Element` and then no cast would be needed here at all.

If changing `findRequiredElement` to have return type `HTMLElement` is not appropriate, I'd add a method `findRequiredHTMLElement` that does what I'm describing instead. In either case, you'd confine all of this casting to a single site.

src/ColumnsPaginatedBookView.ts
```diff
@@ -92,24 +96,46 @@ export default class ColumnsPaginatedBookView implements PaginatedBookView {
...
     /** Returns the total width of the columns that are currently
         positioned to the left of the iframe viewport. */
     private getLeftColumnsWidth(): number {
-        return -(this.bookElement.contentDocument.body.style.left || "0px").slice(0, -2);
+        const body = HTMLUtilities.findRequiredElement(this.bookElement.contentDocument, "body") as HTMLBodyElement;
+
+        const isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') !== -1;
+        const isXML = this.bookElement.src.indexOf(".xml") !== -1;
+        if (isFirefox && isXML) {
+            // This handles books from Feedbooks with XML resources.
```
winniequinn: A little more elaboration here would be nice. I don't understand the need for this special case.

test/ColumnsPaginatedBookViewTests.ts
```diff
@@ -38,7 +38,7 @@ describe("ColumnsPaginatedBookView", () => {
         });
 
         it("should set iframe and iframe body width and height and column width based on window width and set height", () => {
-            (document.body as any).offsetWidth = 100;
+            (window as any).innerWidth = 100;
```
winniequinn, line 41: I'm sure there's a good reason for it, but why are these casts of window to `any` needed? Just curious!

winniequinn (changes requested):

src/ColumnsPaginatedBookView.ts
aslagle: I wish I could do this but HTMLElement doesn't have vendor-prefixed style attributes either. The best I could do would be to cast body.style to any instead of body.

test/ColumnsPaginatedBookViewTests.ts
aslagle: I think window.innerWidth is read-only.

winniequinn: Got it!

src/ColumnsPaginatedBookView.ts
```diff
@@ -157,7 +159,9 @@ export default class ColumnsPaginatedBookView implements PaginatedBookView {
         const isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') !== -1;
         const isXML = this.bookElement.src.indexOf(".xml") !== -1;
         if (isFirefox && isXML) {
-            // This handles books from Feedbooks with XML resources.
+            // Feedbooks epubs have resources with .xml file extensions for historical
+            // reasons. Firefox handles these differently than XHTML files, and setting
+            // a left position as well as overflow:hidden causes the pages to be blank.
```
winniequinn: ❤️

PR #97, opened 2017-05-04
Nav and toc fixes

### LINKED ISSUE #29: Open to the first spine item instead of the TOC

### LINKED ISSUE #75: Chapter display in footer of view
Currently the footer displays "Chapter" instead of "Current Chapter or the TOC or Spine Item name for the the Resource.    It would be more informative to display the aforementioned instead of just "Chapter" next to the page location of the current resource in the footer.

### LINKED ISSUE #81: Add title element to iframe
Add a title element to the iframe containing the book text.

### LINKED ISSUE #82: Chapter navigation control does not disappear
If switching from paged navigation to scroll navigation, then taping (or clicking ) the center of the screen to make the navigation menues disappear, the chapter navigation at the bottom does not disappear.  to make it disappear, the user must tap the center again to make the menues disappear then tap again to make both the menu and the chapter navigation disappear.

### LINKED ISSUE #83: Menu Bar does not always disappears in IOS when scrolling.
img_0983
The menu bard does not always disappear.  When scrolling.  It seems to floats mid screen.

### LINKED ISSUE #95: Book should not be scrollable while TOC is up
This can cause you to lose your place while browsing the TOC.

aslagle:
This fixes #29 
This fixes #75 
This fixes #83 
This fixes #82 
This fixes #81 
This fixes #95 
This fixes https://github.com/NYPL-Simplified/web-reader-landing-page/issues/175

### DIFF
--- a/src/styles/sass/main.scss
+++ b/src/styles/sass/main.scss
@@ -128,11 +128,24 @@ a {
         position: relative;
         text-align: left;
         right: 0;
+        margin: 0 65vw 0 0;
 
         a[rel="up"],
         .icon,
         span, {
           //display: none;
+        }
+
+        @media (min-width: 30rem) {
+          margin-right: 25vw;
+        }
+
+        @media (min-width: 60rem) {
+          margin-right: 60vw;
+        }
+
+        @media (min-width: 87.5rem) {
+          margin-right: 73vw;
         }
       }
 
@@ -140,7 +153,8 @@ a {
         > li:nth-child(1) {
           background-color: $nypl-white;
           display: inline-block;
-          margin: 0.175rem 0 0 0.625rem;
+          margin-top: 0.175rem;
+          margin-left: 0.625rem;
           padding: 0.5rem 0.35rem 0.3rem;
 
           a[rel="up"] {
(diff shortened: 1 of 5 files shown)


src/styles/sass/main.scss
```diff
@@ -128,19 +128,33 @@ a {
...
+          margin-right: 73vw;
+        }
       }
 
       @media screen and (min-width: 30rem) {
         > li:nth-child(1) {
           background-color: $nypl-white;
           display: inline-block;
-          margin: 0.175rem 0 0 0.625rem;
+          margin-top: 0.175rem;
```
ricardoom, line 156: i suspect _zeroing_ the right & bottom values was causing trouble elsewhere?

ricardoom (approved):

src/styles/sass/main.scss
aslagle: I needed the layout to work with or without the contents button since it won't be there for all the books, so I moved it so the back button has a right margin instead of the contents having a left margin. This was part of moving those margins, I think the 0s still get set elsewhere.

PR #105, opened 2017-05-08
Improved behavior when using pinch-to-zoom on iOS.

aslagle:
On android and iOS 9 and lower, we disable zooming with a viewport meta tag. On iOS 10, that's disabled. Based on @winniequinn's suggestions, this makes the reader content work when zoomed in, but ignore swipes and taps. You have to zoom back out before turning the page. 

This will need a lot of cross-browser testing. So far, it seems to work well on iOS 10.3.1 but not as well on iOS 10.2.1, though still better than what happened before.


winniequinn (approved):
This all looks good to me assuming you get your flaky test fixed!

Issue #94, opened 2017-05-03
Add close buttons to table of contents and settings


aslagle:

wlla:
Currently when a menu is expanded, if a user navigates to the end of the menu items, the focus moves out of the menu and to the next element (the book title). As part of this issue, will the user be prevented from navigating past the end of the menu?

ricardoom:
tabbing seems to work pretty well. looks good!

PR #107, opened 2017-05-09
Config object arguments and configurable static file URLs for service worker to cache

### LINKED ISSUE #54: Change class arguments to config objects

### LINKED ISSUE #104: Make static JS and CSS files for ServiceWorker configurable
So they can be minified by an application, for example.

aslagle:
This fixes #54 
This fixes #104 

When I started minifying JS and CSS, I broke the service worker caching because I had hard-coded file names to be cached on install in the service worker itself. I moved all of those to an argument that gets passed into the service worker cacher instead so they can be configurable without having to modify the built service worker file. Since I was adding an argument I decided to go ahead and do #54 as well. 

I don't actually want to merge this into add-xicon-#94, but it's easier to review that way. I'll change this to go to master once that branch is done.

### DIFF
--- a/src/BookSettings.ts
+++ b/src/BookSettings.ts
@@ -46,7 +46,7 @@ export interface BookSettingsConfig {
     /** Store to save the user's selections in. */
     store: Store,
 
-    /** Array of BookView options. */
+    /** Array of BookViews. */
     bookViews: BookView[],
 
     /** Array of font sizes in pixels sorted from smallest to largest. */
--- a/src/BookSettings.ts
+++ b/src/BookSettings.ts
@@ -42,6 +42,20 @@ const checkSvg = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 45 32" pr
 </svg>
 `;
 
+export interface BookSettingsConfig {
+    /** Store to save the user's selections in. */
+    store: Store,
+
+    /** Array of BookView options. */
+    bookViews: BookView[],
+
+    /** Array of font sizes in pixels sorted from smallest to largest. */
+    fontSizesInPixels: number[],
+
+    /** Initial font size to use until the user makes a selection. */
+    defaultFontSizeInPixels?: number;
+}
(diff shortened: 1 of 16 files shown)


src/BookSettings.ts
```diff
@@ -42,6 +42,20 @@ const checkSvg = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 45 32" pr
 </svg>
 `;
 
+export interface BookSettingsConfig {
+    /** Store to save the user's selections in. */
+    store: Store,
+
+    /** Array of BookView options. */
```
winniequinn: In what sense is this an "Array of BookView options"? Did you perhaps mean an "Array of optional BookView objects"? I realize the same comment was in the previous version of this code but I do not understand it there either.

src/IFrameNavigator.ts
```diff
@@ -156,27 +172,15 @@ export default class IFrameNavigator implements Navigator {
...
-
+    public static async create(config: IFrameNavigatorConfig) {
         const navigator = new this(
-            store, cacher, settings, annotator,
-            paginator, scroller, eventHandler,
-            upUrl, upLabel
+            config.store, config.cacher, config.settings, config.annotator || null,
+            config.paginator || null, config.scroller || null, config.eventHandler || null,
+            config.upLink ? config.upLink.url || null : null,
+            config.upLink ? config.upLink.label || null : null
```
winniequinn, line 180: Ow. This is not very pretty but no suggestions immediately jump out at me as being less painful. Blame JavaScript for having `null` _and_ `undefined`, I suppose.

winniequinn (changes requested):

src/BookSettings.ts
aslagle: It is an array of `BookView`s, and the user can select which one to use in settings. They are "options" for the end user. You're right, this comment wasn't a good explanation.

PR #128, opened 2017-05-19
Account for parent margins when setting max image height, and ensure that right column width is never negative.

### LINKED ISSUE #120: Investigate "Page 2 of 1" on Galaxy s6

### LINKED ISSUE #122: Images are still followed by a blank page in some cases

aslagle:
Fixes #120 
Fixes #122

### DIFF
--- a/src/ColumnsPaginatedBookView.ts
+++ b/src/ColumnsPaginatedBookView.ts
@@ -63,7 +63,7 @@ export default class ColumnsPaginatedBookView implements PaginatedBookView {
         // all the vendor-prefixed attributes.
         const body = HTMLUtilities.findRequiredElement(this.bookElement.contentDocument, "body") as any;
 
-        const width = (BrowserUtilities.getWidth() - this.sideMargin * 2) + "px"
+        const width = (BrowserUtilities.getWidth() - this.sideMargin * 2) + "px";
         body.style.columnWidth = width;
         body.style.WebkitColumnWidth = width;
         body.style.MozColumnWidth = width;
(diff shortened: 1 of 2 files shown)


src/ColumnsPaginatedBookView.ts
```diff
@@ -167,7 +181,7 @@ export default class ColumnsPaginatedBookView implements PaginatedBookView {
             // scrollWidth doesn't change when some columns
             // are off to the left, so we need to subtract them.
             const leftWidth = this.getLeftColumnsWidth();
-            rightWidth = rightWidth - leftWidth;
+            rightWidth = Math.max(0, rightWidth - leftWidth);
```
aslagle, line 184: I think this will fix #120.

src/ColumnsPaginatedBookView.ts
```diff
@@ -83,7 +83,21 @@ export default class ColumnsPaginatedBookView implements PaginatedBookView {
...
+                const computedStyle = window.getComputedStyle(nextElement);
+                if (computedStyle.marginTop) {
+                    totalMargins += parseInt(computedStyle.marginTop.slice(0, -2), 10)
+                }
+                if (computedStyle.marginBottom) {
+                    totalMargins += parseInt(computedStyle.marginBottom.slice(0, -2), 10)
+                }
+                nextElement = nextElement.parentElement;
+            }
+            image.style.maxHeight = (this.height - totalMargins) + "px";
```
aslagle, line 100: This is better than before, but it still doesn't fix #122 in all cases. For example. the publisher might put an empty `<p>` tag in front of the image instead of making it a parent. But it's hard to distinguish that from actual content.

aslagle: I opened #129 for that case.

winniequinn (approved):

Issue #136, opened 2017-06-02
Add manifest to HTML


HadrienGardeur:
Now that the JSON-LD context has a proper HTTP redirect (http://readium.org/webpub/default.jsonld) it would be good to inject the manifest in the HTML page for the viewer using `<script type="application/ld+json">` to test if this works properly with search engines.

danielweck:
For the sake of argument, could this JSON-LD link also be added to the "landing page" of each `r2-streamer-js` publication?
For example:
https://readium2.herokuapp.com/pub/L2FwcC9jaGlsZHJlbnMtbGl0ZXJhdHVyZS5lcHVi/
I am not familiar with http://readium.org/webpub/default.jsonld so I don't actually now how to use it :)
Tips welcome (documentation page?).

HadrienGardeur:
The context should be added to each manifest outputted by the streamer.

If you'd like to test, try the example at the top of https://github.com/readium/webpub-manifest

You can try it at:

- https://json-ld.org/playground/
- http://rdf-translator.appspot.com/

I've also published a few outputs (using N3) before at: https://gist.github.com/HadrienGardeur/03ab96f5770b0512233a

danielweck:
Great thanks. I added the `@context` property with URI http://readium.org/webpub/default.jsonld to the generated `manifest.json` in `r2-streamer-js`.

danielweck:
I also added `<script type="application/ld+json" href="URL_manifest.json"></script>` to `html > head` in the "landing" page that `r2-streamer-js` generates to pretty-print the webpub manifest JSON.

PR #137, opened 2017-06-02
Added json manifest script tag to html in examples and added new context url to manifests in examples.

aslagle:
@HadrienGardeur, is this what you meant in #136?


HadrienGardeur:
Yeah looks good. I just need to double check if Google can index external reference to JSON-LD or if it needs to be inline.

PR #139, opened 2018-10-03
JSON `spine` now `readingOrder` in Readium webpub manifest

danielweck:
This code fix preserves `spine` for backward-compatibility.

### DIFF
--- a/src/ServiceWorkerCacher.ts
+++ b/src/ServiceWorkerCacher.ts
@@ -104,7 +104,8 @@ export default class ServiceWorkerCacher implements Cacher {
 
     private async cacheSpine(manifest: Manifest, manifestUrl: URL): Promise<void> {
         const urls: Array<string> = [];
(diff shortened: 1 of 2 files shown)


src/ServiceWorkerCacher.ts
```diff
@@ -104,7 +104,8 @@ export default class ServiceWorkerCacher implements Cacher {
 
     private async cacheSpine(manifest: Manifest, manifestUrl: URL): Promise<void> {
         const urls: Array<string> = [];
-        for (const resource of manifest.spine) {
+        const spn = manifest.readingOrder || manifest.spine;
```
aslagle: Thanks for updating this!

This line isn't necessary and is probably a typescript error since `manifest` here is a `Manifest` instance, not the original JSON.

danielweck:
Thanks @aslagle, my bad!!
I reverted the atomic Git commit.

PR #140, opened 2018-11-21
Update url polyfill

EdwinGuzman:
Updating the npm package version of url-polyfill.

### DIFF
--- a/test/IFrameNavigatorTests.ts
+++ b/test/IFrameNavigatorTests.ts
@@ -515,7 +515,7 @@ describe("IFrameNavigator", () => {
                     ariaLabel: "Up Aria Text"
                 }
             });
-            
+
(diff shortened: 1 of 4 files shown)


test/IFrameNavigatorTests.ts
```diff
@@ -987,15 +987,16 @@ describe("IFrameNavigator", () => {
             expect(link4.text).to.equal("Item 2");
 
             const sublinks = link1.parentElement.querySelectorAll("ul > li > a");
-            expect(sublinks.length).to.equal(2);
+            expect(sublinks.length).to.equal(3);
```
EdwinGuzman, line 990: After updating the url-polyfill package, it seems like the first link is counted as part of the sublinks. When I ran the repo locally, I saw what seemed like a correct TOC.

PR #145, opened 2019-10-01
SFR-599 deploy webpub viewer to elastic beanstalk

hokei:
Added deployment configurations for Elastic Beanstalk

Moved some devDependencies to dependencies so that the app will install in production
See the deployed app here: http://researchnow-webpub-dev.us-east-1.elasticbeanstalk.com:4444/

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,10 +31,15 @@
     "copy:readium-css": "cp -r ./node_modules/readium-css/css/dist examples/streamed/readers/reader-JBKS/readium-css",
     "copy:r2-glue-js": "cp ./node_modules/r2-glue-js/dist/ReadiumGlue.js examples/streamed/readers/reader-JBKS/r2-glue-js.js",
     "copy:r2-navigator": "mkdir -p examples/streamed/readers/reader-JBKS/@readium && cp ./node_modules/@readium/navigator-web/dist/readium-navigator-web.umd.js examples/streamed/readers/reader-JBKS/@readium/navigator-web.js",
-    "copy:streamed": "cpx 'viewer/*.{html,css,js}' examples/streamed/readers/reader-JBKS"
+    "copy:streamed": "cpx 'viewer/*.{html,css,js}' examples/streamed/readers/reader-JBKS",
+    "eb-start": "npm run prepare && npm run examples && npm run streamed"
   },
   "dependencies": {
+    "cpx": "^1.5.0",
+    "node-sass": "^4.11.0",
     "promise-polyfill": "^8.0.0",
+    "r2-streamer-js": "1.0.7",
+    "typescript": "^3.6.3",
     "url-polyfill": "^1.0.13",
     "whatwg-fetch": "^2.0.4"
   },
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -31,10 +31,15 @@
     "copy:readium-css": "cp -r ./node_modules/readium-css/css/dist examples/streamed/readers/reader-JBKS/readium-css",
     "copy:r2-glue-js": "cp ./node_modules/r2-glue-js/dist/ReadiumGlue.js examples/streamed/readers/reader-JBKS/r2-glue-js.js",
     "copy:r2-navigator": "mkdir -p examples/streamed/readers/reader-JBKS/@readium && cp ./node_modules/@readium/navigator-web/dist/readium-navigator-web.umd.js examples/streamed/readers/reader-JBKS/@readium/navigator-web.js",
-    "copy:streamed": "cpx 'viewer/*.{html,css,js}' examples/streamed/readers/reader-JBKS"
+    "copy:streamed": "cpx 'viewer/*.{html,css,js}' examples/streamed/readers/reader-JBKS",
+    "eb-start": "npm run prepare && npm run examples && npm run streamed"
```
EdwinGuzman, line 35: Is this meant to be temporary? In a production env, you wouldn't want examples to run.

EdwinGuzman:
I think the original use of this was to be included into a project and not necessarily be the project itself. So you'd have a compiled version of it imported in a different app. If that can work, then you don't need to move those dev dependencies into the dependencies, which is preferable. I'm not sure what the architecture of RNow is but this won't be included into the main app?

package.json
hokei: So from what I understand of the way webpub-viewer worked, you had to build the examples to be able to run streamed.  I'll look into why this is, I agree it's not necessary to run all the examples.

EdwinGuzman:
I'm going to check later how it was imported into the `opds-web-client` and get back to you.

hokei:
Hmm ... I had been thinking of it as a standalone app this whole time, but a lot of things make a lot more sense if it isn't!  The immediate plan was to replace the epub.js reader with this, one for one, as a temporary solution so that we can give UX an idea of what the app might look like with the new reader.  If it's not a big lift to incorporate it the proper way, though, I'd rather do that!

EdwinGuzman:
I'm not sure which is better and I don't know how the version of this with the most recent updates works. I believe for the Subway Library project we had a static version of it and that's similar to what I had in mind: a compiled or static version of the reader that can do what you want it to. That way, you don't need to download devdependencies into the main project. _But_, if this is meant to be standalone then it should be fine.

What does worry me is the beanstalk configs which are specific to RNow. Someone else using this project may have no need for those files (unless if they want to use it to scaffold their own project on EB).

hokei:
If I understand what's going on, I don't think a static version of the reader is going to work here, because it's running zipped epubs through r2-streamer-js to be able to open and read them.  I could look into bundling all the dependencies together, but I imagine it might get ... complicated?  I don't know if this fact blocks us from incorporating this into RNow right now.  

This isn't meant to be standalone long-term, it's just the fastest way to get RNow to be using this reader.  I'm looking at opds-web-client now and seeing if that approach might work again, though I'm sure you took out webpub-viewer for a good reason and I'm not sure whatever was causing problems has been fixed/I can fix it

I totally see what you mean with the beanstalk configs.  I'm not sure what we can do about them, even though I agree with you that this repo isn't really the place for them.  I'm also a bit wary of things that are intended to be temporary - even though we go in with full intentions to change something, it can stick around for a while.

I think the way to move forward here is to look into whether this can be incorporated into RNow right away.  If it can't, would it work as a temporary solution to push these configs onto another branch, and put something on RNow's roadmap to look into incorporating it as soon as we can?

EdwinGuzman (commented):
The `develop` branch is meant to be the branch for RNow and long-term updates. I definitely don't want the beanstalk configs to go into `master` but it's fine if it goes into `develop`. There should be something on the roadmap that says that a better longer-term solution will be worked on. I think having those in the repo will make it less generic. I would just write some comments to note that some change are temporary.

The opds-web-client integration stopped working because of some dependencies but it may work now/soon.

EdwinGuzman (approved):

PR #146, opened 2019-12-04
SFR-760_make_text_selectable

hokei:
A note:  the `.css` and the `.js` files are generated.  It would be better to put them in a gitignore.

### DIFF
--- a/src/IFrameNavigator.ts
+++ b/src/IFrameNavigator.ts
@@ -1023,12 +1023,6 @@ export default class IFrameNavigator implements Navigator {
         body.style.fontSize = fontSize;
         body.style.lineHeight = "1.5";
 
-        // Disable text selection as we can’t handle this correctly anyway
-        body.style.webkitUserSelect = "none";
-        (body as any).style.MozUserSelect = "none";
-        body.style.msUserSelect = "none";
-        body.style.userSelect = "none";
-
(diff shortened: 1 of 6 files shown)


src/IFrameNavigator.ts
```diff
@@ -1023,12 +1023,6 @@ export default class IFrameNavigator implements Navigator {
         body.style.fontSize = fontSize;
         body.style.lineHeight = "1.5";
 
-        // Disable text selection as we can’t handle this correctly anyway
```
mwbenowitz, line 1026: Do we know what the original reason for this was / what "can't handle this correctly" meant? Not that I think this is an issue just wondering because perhaps there is a specific bug that we should keep an eye out for. (maybe specific to mobile devices?)

mwbenowitz (approved):
Looks good to me, just wondering what the original reason for this being disabled is, don't think that it should prevent us from deploying the fix though.

src/IFrameNavigator.ts
seanredmond: The commit that added it was to "Make status messages non-selectable", so maybe it was overkill?

mwbenowitz: hmm, yeah, if that's the case it would seem that way to me

hokei: A good thing to keep an eye out for anyway!

PR #147, opened 2019-12-13
SFR-770 added NYPL logo and changed back-links to events

hokei:
I want to rename the develop branch to a research-now branch, so that it's clear what's happening here, especially since we're going to be making very specific RN changes.  

To test out if this works, pull this guy as well:  https://github.com/NYPL/sfr-bookfinder-front-end/tree/SFR-770_webpub-back-button


mwbenowitz (commented):
This looks way better!

Two notes:

- When I click through the book the back button first takes me back through the pages I viewed before returning me to my search result. It does work but is there a way to "jump" through the history array to get to the last non-viewer link? I think this is related to the react app so I'll leave a comment there as well
- The "Return to ResearchNow" text is read as text and displays a text cursor when it is hovered over. This isn't a problem but it would be nice if it registered as a button and displayed a pointer as a hint to the user.

PR #151, opened 2020-03-16
SFR-831 use nypl reader

hokei:
Changes Streamed example to use a subclass of r2-streamer-js' `server.js` such that it is not offering non-NYPL readers, as this causes confusion.


README.md
```diff
@@ -76,9 +76,7 @@ For the origin to be considered the same, protocol (`http/https`), host and port
 
 ### Streamed
 
-The `examples/streamed` folder contains the webpub-viewer (`misc/readers/reader-NYPL` folder) and example files (`epubs` folder).
-
-In this example, assets are served by the r2-streamer-js (in-memory model). You can open them using different reader implementations. This iteration of webpub-viewer is labelled as "Reader NYPL".
+In this example, assets are served by the r2-streamer-js (in-memory model). his iteration of webpub-viewer is labelled as "Reader NYPL".
```
EdwinGuzman: Small typo "this".

EdwinGuzman (commented):
We briefly chatted about this and although this seems right, I just don't think this should be in this repo. I think this repo should still have the examples and work as a package, and then you'd have a different repo that is the "Reader NYPL" app that pulls this package in and serves the streamer. Unless if this really is the direction that the project is taking.

hokei:
We chatted a bit again - Still trying to figure out why this isn't working on `static`.  In the meantime I've updated the readmes so hopefully things can be a bit more clear.

mwbenowitz (approved):

PR #152, opened 2020-03-17
SFR-869 Update Project Name

mwbenowitz:
Adds the new `Digital Research Books Beta` name to the navigation bar in the header to align with the rest of the SFR app.


hokei (approved):
pending travis

PR #158, opened 2020-06-12
Export as ES6 module

hokei:
This should allow the viewer to be exported and imported as an ES6 module.


M0nica (approved):
This looks good to me 👍

PR #161, opened 2020-06-24
Fix Static Loading

M0nica:
There was an issue where the build files for the static viewer were no longer being built properly when changes were made to the src files. 

This PR:
- Updates static express server to try to resolve files without an extension to .js
- Updates module imports and exports as needed 
- Switches module type from common -> es6

Test:
1. Build the project locally - `npm run build`
2. Run static server - `npm run static`
3. Click on "The Call of the Wild" link on index page and the e-reader loads. If you make changes to `.ts` files in the `src` directory or to the index.html in the viewer folder it should effect the appearance of this e-reader (post re-build)


hokei (approved):
🎉

PR #162, opened 2020-06-29
Display Exploded EPubs by parsing container.opf

M0nica:
- Added EPub to parse .opf files and extract metadata, links, spine, resources, toc and manifestUrl for IFrameNavigator to navigate through the e-book.
- This supports the OEBPS (Open Ebook Forum Publication Structure) https://www.loc.gov/preservation/digital/formats/fdd/fdd000171.shtml 

Note: This doesn't handle loading EPubs /container.opf's from external sources. This works with locally hosted EPubs (i.e., in `static` example).


src/EPub.ts
```diff
@@ -0,0 +1,244 @@
...
+        });
+        return acc;
+      },
+      []
+    );
+  }
+  public constructor(manifestJSON: any, manifestUrl: URL) {
+    this.metadata = this.parseMetaData(manifestJSON);
+    //links format should be updated to point to manifest.json
+    this.links = this.parseTOC(manifestJSON);
```
M0nica:
```
{
        href: manifestJSON.href,
        type: "application/webpub+json",
        templated: false,
        rel: "self",
      }
``` 

`this.links` should look something like the above but it's supposed to point to a _manifest.json_ file. I am not sure if links are optional for the e-reader?

test/EPubTests.ts
```diff
@@ -0,0 +1,452 @@
...
+      expect(emptyManifest.links).to.deep.equal([]);
+      expect(emptyManifest.spine).to.deep.equal([]);
+      expect(emptyManifest.resources).to.deep.equal([]);
+    });
+
+    it("should store metadata", () => {
+      expect(manifest.metadata.title).to.equal("The Elephant");
+    });
+
+    it.skip("should store links", () => {
```
M0nica: TBD on if links is necessary for the EPub.ts files

test/EPubTests.ts
```diff
@@ -0,0 +1,452 @@
...
+            store
+          );
+        } catch (err) {
+          rejected = true;
+        }
+        expect(rejected).to.be.true;
+      });
+    });
+
+    it.skip("should return the response from fetch, and save it to local store", async () => {
```
M0nica: I modified this test suite from Manifest.ts but need to understand it better. I am not sure why xx is returning the manifest but saying that an empty response was expected 🤔 


AssertionError: expected { Object (metadata, links, ...) } to deeply equal { Object (metadata, links, ...) }
  ```
      + expected - actual

       {
         "links": []
         "manifestUrl": {}
      -  "metadata": {
      -    "title": "The Elephant"
      -  }
      +  "metadata": {}
         "resources": []
         "spine": []
         "toc": []
       }
```

 I am not sure why it's expecting an empty object, I am expecting ..

If I switch expected to  hard-coded 
```{
          links: [],
          manifestUrl: {},
          metadata: {
            title: "The Elephant",
          },
          resources: [],
          spine: [],
          toc: [],
        }
```

 it fails with `AssertionError: expected { Object (metadata, links, ...) } to deeply equal {}

src/EPub.ts
```diff
@@ -0,0 +1,244 @@
...
+      }
+    }
+
+    return fetchManifest();
+  }
+
+  public parseMetaData(manifestJSON: any): any {
+    return manifestJSON?.package?.metadata
+      ? {
+          title: manifestJSON.package.metadata["dc:title"]["#text"],
```
M0nica: This metadata assignment needs to be updated to dynamically parse more robust metadata from the package.opf:  
```

dc:creator | <empty> or <unknown> | author
-- | -- | --
dc:contributor | <empty> or <unknown> | contributor
dc:creator or dc:contributor | aut | author
dc:creator or dc:contributor | pbl | publisher
dc:creator or dc:contributor | trl | translator
dc:creator or dc:contributor | edt | editor
dc:creator or dc:contributor | ill | illustrator
dc:creator or dc:contributor | art | artist
dc:creator or dc:contributor | clr | colorist
dc:creator or dc:contributor | nrt | narrator
dc:publisher | N/A | publisher
media:narrator | N/A | narrator
```
source: https://github.com/readium/architecture/blob/master/streamer/parser/metadata.md 

Does the e-reader only care about a subset of potential metadata?

hokei: It should parse all of the metadata, if available.  We shouldn't strip it out ...

hokei: Unless this is used to explicitly populate a title or something ... (Where is it used?)

hokei: It looks like the only place that `Manifest.metadata` is used is to set the title in `IFrameNavigator.ts#L899`

In that case you might want to replace `Manifest.metadata` with `Manifest.title` for clarity and save some work

src/EPub.ts
hokei: Aren't the links the chapter links for the TOC?  That seems like an important thing ... but I might be mistaken

src/EPub.ts
```diff
@@ -0,0 +1,244 @@
+import Store from "./Store";
+
+type xmlObject = {
+  [key: string]: string[] | string | xmlObject | [];
+};
+
+function xmlToJson(xml: any) {
```
hokei: does DomParser not work here? (https://developer.mozilla.org/en-US/docs/Web/API/DOMParser)

src/EPub.ts
```diff
@@ -0,0 +1,244 @@
...
+
+export interface Link {
+  rel?: Array<string>;
+  href?: string;
+  type?: string;
+  title?: string;
+  children?: Array<Link>;
+}
+
+export default class Manifest {
```
hokei: Rename classname so it doesn't clash with `Manifest.ts`

src/EPub.ts
M0nica: @hokei I initially thought links was another method for parsing TOC but the standard is for link to contain links to the manifest.json. That's how the other EPub in the static folder is formatting it's link (linking to manifest.json)

The spec is viewable here: https://github.com/readium/webpub-manifest#23-links 
example: 
```
"links": [
    {"rel": "self", "href": "https://example.com/manifest.json", "type": "application/webpub+json"},
    {"rel": "alternate", "href": "https://example.com/publication.epub", "type": "application/epub+zip"},
    {"rel": "search", "href": "https://example.com/search{?query}", "type": "text/html", "templated": true}
  ]
```

src/EPub.ts
```diff
@@ -0,0 +1,244 @@
...
+        .then((data) => JSON.stringify(xmlToJson(data)));
+
+      if (store) {
+        await store.set("manifest", JSON.stringify(manifestJSON));
+      }
+
+      return new Manifest(JSON.parse(manifestJSON), manifestUrl);
+    };
+
+    const tryToUpdateManifestButIgnoreResult = async (): Promise<void> => {
```
hokei: Some of these methods are exactly the same as `Manifest.ts`
I wonder if there's a way to either rework `Manifest.ts` to recognize if it's an epub or not, or extract similar functions, or find some other way to consolidate this code so we don't have two copies of very similar code.  It'll also make it more obvious which parts of this are specific to the webpub stuff and which are specific to epub.

M0nica: I updated to have all of the code in one file for parsing manifest.json and package.opf in the same file. I wonder if it'll still be confusing to refer to both formats as Manifest

src/EPub.ts
M0nica: This function is transforming the output from DomParser into JSON. 

 https://github.com/NYPL-Simplified/webpub-viewer/blob/6a4325c2ee3c836e0a9af413ef1f25640121e8e3/src/Manifest.ts#L85-L88

It looks like the MIME types that are supported by DomParser are: 

text/html | Document
-- | --
text/xml | XMLDocument
application/xml | XMLDocument
application/xhtml+xml | XMLDocument
image/svg+xml | XMLDocument

src/EPub.ts
hokei: yeah definitely, it's not really "manifest" as it is some kind of ... parser for the manifest-type file?  Aaaah names are hard.

M0nica: I renamed the parsed `package.opf` to `OPFPackage` in certain places for now.

examples/streamed/streamed-server.js
```diff
@@ -1,54 +1,68 @@
...
-    this.disableReaders=true;
+    this.disableReaders = true;
     this.publications = [];
     this.pathPublicationMap = {};
     this.expressApp = express();
 
-    this.expressApp.use("/readerNYPL", express.static(__dirname + "/../../viewer"));
+    this.expressApp.use(
+      "/readerNYPL",
+      express.static(__dirname + "/../../viewer", { extensions: ["js"] })
```
M0nica, line 40: This change here makes is to so the express will look for missing files with the `.js` extension so if it’s trying to get
http://localhost:4444/readerNYPL/HTMLUtilities and it 404s
it should check for `
http://localhost:4444/readerNYPL/HTMLUtilities.js` and only return a 404 if that is not found

examples/streamed/streamed-server.js
```diff
@@ -1,54 +1,68 @@
-var express = require('express');
+var express = require("express");
```
hokei, line 1: if it's not much trouble can you commit your vscode settings as well so we can stay in sync?  I'm not sure I'm using anything, as evidenced by the inconsistent nature of this file.

hokei:
can you add everything under `/viewer` to gitignore?  I don't think there's a reason it should be committed ...

examples/streamed/streamed-server.js
M0nica: I added `.prettierrc.json` with my configuration

M0nica:
> can you add everything under `/viewer` to gitignore? I don't think there's a reason it should be committed ...

I updated to add all of the files in `/viewer` aside from index.html to the .gitignore and removed all of the `/viewer` files from git.

src/Manifest.ts
```diff
@@ -1,135 +1,284 @@
...
-            }
-        }
+  public static async getManifest(
+    manifestUrl: URL,
+    store?: Store
+  ): Promise<Manifest> {
+    const fetchManifest = async (): Promise<Manifest> => {
+      const isJSONManifest = Boolean(manifestUrl.href.endsWith(".json"));
+
+      const manifestJSON = isJSONManifest
```
hokei: this could use some comments to make sure we know what is happening?  Is this method only for json manifests?

src/Manifest.ts
```diff
@@ -1,135 +1,284 @@
...
-    title?: string;
-    children?: Array<Link>;
+  rel?: Array<string>;
+  href?: string;
+  type?: string;
+  title?: string;
+  children?: Array<Link>;
 }
 
 export default class Manifest {
```
hokei, line 64: Can you put a comment here on what exactly a Manifest is in this context (making it clear that it can be a Package Document or a Webpub Manifest?)

src/Manifest.ts
```diff
@@ -1,135 +1,284 @@
...
+    let OPF = null;
+    try {
+      OPF = JSON.parse(OPFPackage);
+    } catch (e) {
+      OPF = JSON.parse(JSON.stringify(OPFPackage));
     }
+    return OPF;
+  }
+
+  public parseMetaData(OPFPackage: any): any {
```
hokei: Can we have a block comment making clear that there's a whole block of OPF-specific functions?

hokei (approved):
Looks good, I think it would be helpful to have some comments to help us parse this in the future.

PR #163, opened 2020-07-09
Store and load EPub from local storage

M0nica:
- Add  support to extract Manifest information from remote container.xml and *.opf urls.
- Save exploded EPubs to local storage and render in webpub-viewer from local storage in order to render remote EPubs in IFrame without CORS issues.
- Images are rendered by rewriting XHTML files in localstorage to use base64 encoded version from local storage.
- Temporarily CSS is being pulled from remote URL. This, in addition to asset handling will be revisited in future work.

It pulled in all the commits from https://github.com/NYPL-Simplified/webpub-viewer/tree/testingExplodedEPubs since I based this branch off of that one 🤯 

Screenshot of how localhost files are saved in localstorage (it's the same structure for the remote files as well but the key has the remote URL instead)
Screen Shot 2020-07-09 at 2 23 49 PM


src/IFrameNavigator.ts
```diff
@@ -1441,10 +1451,23 @@ export default class IFrameNavigator implements Navigator {
     }
   }
 
-  private navigate(readingPosition: ReadingPosition): void {
+  private async navigate(readingPosition: ReadingPosition): Promise<void> {
     this.hideIframeContents();
     this.showLoadingMessageAfterDelay();
     this.newPosition = readingPosition;
+
+    const baseUrl = this.manifestUrl.href.replace(/[a-z]+.opf/, "");
```
M0nica, line 1490: I'm not sure if this is too hacky to use to shorten from the full path of the resource URL to the short path for that specific file. It assumes the manifest URL for exploded EPubs will be formatted like www.example.com/{library}/{bookname}/{OEBPS}/{packageFile}.opf (`{}` between the brackets can be anything) or 
www.example.com/{...}/{packageFile}.opf and that the resource files will be in the same directory as the packageFile (i.e., OEBPS).

src/Manifest.ts
```diff
@@ -90,8 +118,24 @@ export default class Manifest {
             .then((data) => JSON.stringify(xmlToJson(data)));
 
       if (store) {
+        const resources = parseOPFResources(parseOPFPackage(manifest));
+
+        for (let resource of resources) {
+          const fullResourceUrl = `${manifestUrl.href.replace(
```
M0nica, line 136: Similar concerns as the other version of this RegEx.

src/IFrameNavigator.ts
hokei: I think we can assume that the resources are in the same directory as the manifest.  However, I think we shouldn't assume anything about the structure of the manifest url.  Like, it could be www.example.com/stuff/morestuff/morestuff/morestuff/package.opf

hokei: Scratch that -- I think we ... might need to read all of this using an epub `container.xml` since that's guaranteed to tell us what we need
http://idpf.org/epub/301/spec/epub-ocf.html#sec-container-metainf-container.xml

So I think it would be more correct to pass the viewer either 1) a link to the manifest.json (for webpubs) or 2) a link to the container.xml (for epubs)

hokei (approved):
This is super cool stuff!

src/IFrameNavigator.ts
M0nica: Support for passing container.xml in addition to .ops file has been added

PR #164, opened 2020-07-10
Access Control

hokei:
- Stores book resources in `IndexedDB` instead of `LocalStorage` 
- Adds decryption logic for epubs that use encryption 
- Fixes CSS loading
- Changes `manifestUrl` to be `entryURL` to avoid the confusion between webpub manifests, epub `container.xml` files and epub `content.opf` files.  

TODO:  
- `BookResourceStore` tests
- `Encryption` tests
- Remove all the console.logs


viewer/index.html
```diff
@@ -61,6 +61,7 @@
       var cacher = new ServiceWorkerCacher({
         store: store,
         manifestUrl: webpubManifestUrl,
+        encryptionUrl: new URL("https://www.crummy.com/axis/catcher/encrypted/META-INF/encryption.xml"),
```
hokei: TODO: remove

src/Encryption.ts
```diff
@@ -0,0 +1,53 @@
...
+      }
+    }
+
+    return fetchEncryption();
+  }
+
+  private getEncryptedResourceList(encryptionString: any) {
+    let encryptionData = JSON.parse(encryptionString);
+    return encryptionData.encryption["enc:EncryptedData"].map((data: { [x: string]: { [x: string]: { [x: string]: any; }; }; }) => {
+      return data["enc:CipherData"]["enc:CipherReference"]["@attributes"].URI;
```
hokei, line 46: Is this OK to assume?

src/Container.ts
```diff
@@ -0,0 +1,20 @@
+export default class Container {
```
hokei: TODO: remove

src/IFrameNavigator.ts
```diff
@@ -1442,9 +1454,19 @@ export default class IFrameNavigator implements Navigator {
   }
 
   private navigate(readingPosition: ReadingPosition): void {
+    console.log("readingPosition", readingPosition);
     this.hideIframeContents();
     this.showLoadingMessageAfterDelay();
     this.newPosition = readingPosition;
+    if(this.encryption) {
+      let isEncrypted = this.encryption.resources.find((resource: string) => {
+        return readingPosition.resource.includes(resource);
```
hokei: In the sample that I was using, resources were of the address `opf/chapter1_part1.xhtml` whereas readingPositions were of the address `www.host.com/blah/dee/blah/opf/chapter1_part1.xhtml`

Do we know if this will hold for everything?

src/Encryption.ts
leonardr: For now (and for AxisNow books), I think so -- the Axisnow encryption document is actually slightly wrong somewhere around here, so an assumption might be necessary anyway.

src/IFrameNavigator.ts
leonardr: I don't know; what is a readingPosition in this context? Like which chapter you're currently reading?

examples/static/static-server.js
```diff
@@ -21,26 +22,49 @@ server.listen(port, (err) => {
   }
 });
 
+// const apiProxy = createProxyMiddleware("/axis", {
```
M0nica: I'd revert the changes to the static-server.js. I wonder if it would be helpful to have some type of env variable to use to switch between different versions (i.e, different books) of the static-server.js

src/Encryption.ts
```diff
@@ -0,0 +1,53 @@
...
+    encryptionUrl: URL,
+    store?: Store
+  ): Promise<Encryption> {
+    const fetchEncryption = async (): Promise<Encryption> => {
+        console.log("encryptionUrl", encryptionUrl);
+      const encryption = await window 
+            .fetch(encryptionUrl.href)
+            .then((response) => response.text())
+            .then((str) => 
+              new DOMParser().parseFromString(str, "text/xml")
```
M0nica, line 21: Does this work for `manifest.json`? I'd think that `.json` and `.ops` files would need to be processed separately.

src/Encryption.ts
```diff
@@ -0,0 +1,53 @@
...
+      
+      if (store) {
+        await store.set("encryption", JSON.stringify(encryption));
+      }
+      return new Encryption(JSON.parse(JSON.stringify(encryption)), encryptionUrl);
+    };
+
+    // Respond immediately with the encryption from the store, if possible.
+    if (store) {
+      // const encryptionString = await store.get("encryption");
```
M0nica: Is this store.get functional? or is it commented out because it's not yet functional?

src/IFrameNavigator.ts
hokei: Yeah, the link to the chapter.  `readingPosition.resource` is the direct link to the chapter xhtml.

src/Decryptor.ts
```diff
@@ -0,0 +1,5 @@
+export default interface Decryptor {
```
hokei: This is the equivalent of https://github.com/readium/web/blob/028dfff7b9a0b34f529504cd039f0bac411cb045/shared-models/src/Publication/Encryption/Encryption.ts 

which makes me think that it does belong here, but I need to do another pass on the actual methods to make sure it all makes sense.  Ignore this file for now, I think it goes best with the PR in `axisnow-access-control-web`

examples/static/static-server.js
M0nica: I still think these changes can be reverted. Although, it would be nice to update the static server with an example of `application/vnd.librarysimplified.web-epub`that can be committed to GitHub since right now the `/catcherRye` can't be added to version control.

src/BookResourceStore.ts
```diff
@@ -0,0 +1,97 @@
...
+  private readonly db: IDBDatabase;
+
+  public constructor(db: IDBDatabase) {
+    this.db = db;
+  }
+
+  public static createBookResourceStore(): Promise<BookResourceStore> {
+    if (!("indexedDB" in window)) {
+      //TODO: a polyfill to handle lack of IndexedDB
+      console.log("This browser doesn't support IndexedDB");
```
M0nica, line 20: If there is not `"indexedDB" in window`should the code skip line 22? https://github.com/NYPL-Simplified/webpub-viewer/pull/164/files#diff-0cdd1f0c6fc8a7bc70f13c6c6a321f94R22

src/BookResourceStore.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+    var request = window.indexedDB.open("WebpubViewerDb", 2);
+    let store:BookResourceStore;
+    return new Promise((resolve) => {
+      request.onupgradeneeded = (evt) => {
+        let db = (<any>evt.target).result;
+        if (!db.objectStoreNames.contains("bookResources")) {
+          var bookResourceOS = db.createObjectStore("bookResources", {
```
M0nica: I think `bookResourceStore` might be more clear than `bookResourceOS`

src/BookResourceStore.ts
```diff
@@ -0,0 +1,97 @@
...
+          });
+          bookResourceOS.transaction.oncomplete = () => {
+            store = new BookResourceStore(db);
+          }
+        }
+      }
+
+      request.onsuccess = (evt) => {
+        if(!store) {
+          store = new BookResourceStore((<any>evt.target).result);
```
M0nica, line 39: Can some of these IndexedDB related `any`s be typed?

src/BookResourceStore.ts
```diff
@@ -0,0 +1,97 @@
...
+    });
+  }
+
+  addAllBookData(manifest: Manifest) {
+    return Promise.all(manifest.resources.map(async (resource: any) => {
+      const manifestPath = manifest.manifestUrl.href.substring(
+        0,
+        manifest.manifestUrl.href.lastIndexOf("/")
+      );
+      //add leading slash to resource.href if not exists
```
M0nica, line 87: I would've missed the difference between the two conditions if it wasn't for this comment.

M0nica (commented):
overall it looks good to me so far

src/Encryption.ts
hokei: `manifest.json` files are assumed to be not encrypted - we don't have an example of an encrypted one, and I imagine the AxisNow decryptor would not work with them anyway.  I should add something to the readme about this ...

src/BookResourceStore.ts
hokei: There is nothing yet to handle what happens when indexedDB doesn't exist ... 
I filed https://jira.nypl.org/browse/SIMPLY-2914 but it might be good to just have it always load from the remote

examples/static/static-server.js
```diff
@@ -28,13 +29,14 @@ app.use(
 );
 app.use("/backstop", express.static(__dirname + "/backstop"));
 
+app.use("/catcherRye", express.static(__dirname + "/catcherRye"));
```
hokei: remove

examples/static/static-server.js
```diff
@@ -1,6 +1,7 @@
 const https = require("https");
 const fs = require("fs");
 const express = require("express");
+// const { createProxyMiddleware } = require("http-proxy-middleware");
```
M0nica: 🔪

src/BookResourceStore.ts
hokei: Theoretically, yes.  Practically, I can't figure out what the type name for `IDBVersionChangeEvent.Target` is ... because `IDBVersionChangeEvent` doesn't have a target value documented.  Clearly it exists, but I can't find any mention of it in the documentation

src/BookResourceStore.ts
```diff
@@ -0,0 +1,99 @@
...
+export default class BookResourceStore {
+  private readonly db: IDBDatabase;
+
+  public constructor(db: IDBDatabase) {
+    this.db = db;
+  }
+
+  public static createBookResourceStore(): Promise<BookResourceStore> {
+    if (!("indexedDB" in window)) {
+      //TODO: a polyfill to handle lack of IndexedDB
```
M0nica, line 19: Can you make sure a ticket gets create to handle polyfilling for browsers that don't have IndexedDB?

src/IFrameNavigator.ts
```diff
@@ -1502,28 +1527,72 @@ export default class IFrameNavigator implements Navigator {
...
+
   private async navigate(readingPosition: ReadingPosition): Promise<void> {
+    console.log("navigate called");
     this.hideIframeContents();
     this.showLoadingMessageAfterDelay();
     this.newPosition = readingPosition;
 
-    const baseUrl = this.manifestUrl.href.replace(/[a-z]+.opf/, "");
-    const shortResourceUrl = readingPosition.resource.replace(baseUrl, "");
+    console.log("readingposition aaaaa resource", readingPosition.resource);
```
M0nica: There are some extra `console.log`s in this file that should be removed.

src/LocalAnnotator.ts
```diff
@@ -16,6 +16,7 @@ export default class LocalAnnotator implements Annotator {
 
     public async getLastReadingPosition(): Promise<any> {
         const positionString = await this.store.get(LocalAnnotator.LAST_READING_POSITION);
+        console.log("positionString", positionString);
```
M0nica: 🔪

src/Encryption.ts
```diff
@@ -0,0 +1,75 @@
...
+            .then((data) => JSON.stringify(Utils.xmlToJson(data)));
+      
+      if (store) {
+        await store.set("encryption", JSON.stringify(encryption));
+      }
+      return new Encryption(JSON.parse(JSON.stringify(encryption)), encryptionUrl);
+    };
+
+    // Respond immediately with the encryption from the store, if possible.
+    if (store) {
```
hokei, line 33: There's a bug here -- things work the first time a book is loaded but not on refresh.  Comment this if block out until I figure out what's going on

M0nica (approved):

PR #166, opened 2020-07-20
Allow custom library icons to be displayed in nav

M0nica:
Currently, the NYPL Logo is hard-coded within the uplink (nav) portion of the webpub-viewer. The uplink already has customizable url, label, and ariaLabel.

This PR updates the webpub-viewer to allow passing in a custom logo (i.e., Open E-Books logo) URL to display in nav. The default icon will be the NYPL icon so this should not be a broken change.

### DIFF
--- a/src/IFrameNavigator.ts
+++ b/src/IFrameNavigator.ts
@@ -37,9 +37,10 @@ const upLinkImage = (label: string, image?: string) => {
 const upLinkTemplate = (
   label: string,
   ariaLabel: string,
-  libraryIcon: string
+  libraryIcon: string,
+  url: string
 ) => `
-  <a rel="up" aria-label="${ariaLabel}" tabindex="0">
+  <a href="${url}" rel="up" aria-label="${ariaLabel}" tabindex="0">
   ${upLinkImage(label, libraryIcon)}
     <span class="setting-text up">${label}</span>
   </a>
@@ -791,10 +792,16 @@ export default class IFrameNavigator implements Navigator {
       }
 
       if (this.upLinkConfig && this.upLinkConfig.url) {
+        const upUrl = this.upLinkConfig.url.href;
         const upLabel = this.upLinkConfig.label || "";
         const upAriaLabel = this.upLinkConfig.ariaLabel || upLabel;
         const upLibraryIcon = this.upLinkConfig.libraryIcon?.href || "";
-        const upHTML = upLinkTemplate(upLabel, upAriaLabel, upLibraryIcon);
+        const upHTML = upLinkTemplate(
+          upLabel,
+          upAriaLabel,
+          upLibraryIcon,
+          upUrl
+        );
         const upParent: HTMLLIElement = document.createElement("li");
         upParent.classList.add("uplink-wrapper");
         upParent.innerHTML = upHTML;
--- a/src/IFrameNavigator.ts
+++ b/src/IFrameNavigator.ts
@@ -147,10 +147,10 @@ interface ReadingPosition {
 }
 
 export interface UpLinkConfig {
-  url?: URL;
-  label?: string;
-  ariaLabel?: string;
-  libraryIcon?: URL;
+  url: URL;
+  label: string;
+  ariaLabel: string;
+  libraryIcon: URL;
 }
 
 export interface IFrameNavigatorConfig {
--- a/src/IFrameNavigator.ts
+++ b/src/IFrameNavigator.ts
@@ -37,8 +37,8 @@ const upLinkImage = (label: string, image?: string) => {
 const upLinkTemplate = (
   label: string,
   ariaLabel: string,
-  libraryIcon: string,
-  url: string
+  url: string,
+  libraryIcon?: string
 ) => `
   <a href="${url}" rel="up" aria-label="${ariaLabel}" tabindex="0">
   ${upLinkImage(label, libraryIcon)}
@@ -151,7 +151,7 @@ export interface UpLinkConfig {
   url: URL;
   label: string;
   ariaLabel: string;
-  libraryIcon: URL;
+  libraryIcon?: URL;
 }
 
 export interface IFrameNavigatorConfig {
@@ -799,8 +799,8 @@ export default class IFrameNavigator implements Navigator {
         const upHTML = upLinkTemplate(
           upLabel,
           upAriaLabel,
-          upLibraryIcon,
-          upUrl
+          upUrl,
+          upLibraryIcon
         );
         const upParent: HTMLLIElement = document.createElement("li");
         upParent.classList.add("uplink-wrapper");
--- a/src/IFrameNavigator.ts
+++ b/src/IFrameNavigator.ts
@@ -24,12 +24,19 @@ const epubReadingSystemObject: EpubReadingSystemObject = {
 
 const epubReadingSystem = Object.freeze(epubReadingSystemObject);
 
-const upLinkTemplate = (label: string, ariaLabel: string) => `
+const upLinkImage = (label: string, image?: string) => {
+  if (image) {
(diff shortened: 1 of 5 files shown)


hokei (changes requested):
Is there a unit test that makes sense here?

src/IFrameNavigator.ts
```diff
@@ -139,6 +146,7 @@ export interface UpLinkConfig {
   url?: URL;
   label?: string;
   ariaLabel?: string;
+  libraryIcon?: string;
```
hokei: Maybe change this to "logo" since webpub-viewer isn't always going to be used by libraries?

hokei: Also, should this be a URL instead?

M0nica: The `logoUrl` in `circulation-patron-web` is defined as a `string` but I can update web-pub viewer to expect a URL and just wrap the logoUrl in a new URL. https://github.com/NYPL-Simplified/circulation-patron-web/blob/7b73e4640f619616b2cfb5d649232ef0c7297b88/src/interfaces.ts#L50 

a URL will hopefully make it clear that SVG can't be passed in!

M0nica: I updated for the libraryIcon type to be URL instead of string.

M0nica:
> Is there a unit test that makes sense here?

I added a couple of unit tests:
- check that the image url for the custom image is displayed when custom image is passed in AND that SVG does not appear
- check that default icon is displayed even if an custom library icon is not passed in AND that there is no `src` attribute

I am not sure if there's a better way to check whether or not an svg is rendered, right now I am using the presence/absence of "aria-labelled-by" since the svg icon uses that attribute but the image attributes use the "title" attribute directly. If you have suggestions for a better selector I can update the tests!

test/IFrameNavigatorTests.ts
```diff
@@ -657,6 +657,71 @@ describe("IFrameNavigator", () => {
       });
     });
 
+    it("should show the default SVG logo when custom LibraryIcon is not passed in", async () => {
```
hokei: This is weird.  I think instead of making libraryIcon optional. let's make upLink optional.  Chances are, you either want to pass a custom logo/link or you don't, I can't imagine that anyone wants to pass in a link to their own site but keep the NYPL logo.  So the tests can just check for, "if you pass a new upLink in, it shows up" and "if you don't, then the NYPL link and logo shows up" 

(It doesn't matter what the NYPL link is, in non-drb it was linking to this github repo, I think that's ok)

M0nica: @hokei  what should the default NYPL link, title and aria-label information to be if someone doesn’t pass in an uplink? Would it be better to hide the uplink if all of the information isn't provided? Currently it's only checking that there is an upLinkConfig with a url.

Currently, all of the items in UpLinkConfig are optional https://github.com/NYPL-Simplified/webpub-viewer/blob/d16ce729d9ab127eec7884df11fada09b108edd8/src/IFrameNavigator.ts#L149-L154 are you recommending that url, label, ariaLabel and libraryIcon become required while upLink remains optional

hokei: Oh I was thinking https://github.com/NYPL-Simplified/webpub-viewer/blob/d16ce729d9ab127eec7884df11fada09b108edd8/src/IFrameNavigator.ts#L172 but it seems like it's already optional.  
I think you either have your own configuration, or you don't ... and if you have your own configuration, url/label/ariaLabel/libraryIcon should be required.  (Maybe url can be optional, but honestly I don't think it's a big deal for the person who needs the words to not be a link to make that change when they get to it)  

Does that make sense?

M0nica: Yep, that makes sense 👍 I will update accordingly.

viewer/index.html
```diff
@@ -100,6 +100,7 @@
                 url: new URL("https://github.com/NYPL-Simplified/webpub-viewer"),
                 label: "Return to Digital Research Books Beta",
                 ariaLabel: "Return to Digital Research Books  Beta",
+                libraryIcon: new URL("https://www.nypl.org/sites/default/files/images/av/news_2009_11_06_logo.jpeg")
```
hokei: It's weird to me that you now have to pass in a jpeg for the library logo?  If it defaults to the svg when nothing is passed in, then let's just not pass anything in.  It's either that, or allow the libraryIcon to be passed as an svg.

hokei (changes requested):
Sorry for always having tiny things ... It looks pretty good to me, just that question about the library icon jpeg

test/IFrameNavigatorTests.ts
```diff
@@ -657,6 +657,103 @@ describe("IFrameNavigator", () => {
       });
     });
 
+    it("should show the default SVG logo when custom LibraryIcon is an empty string", async () => {
```
hokei: So based on these tests, it still seems to work really strangely to me.  1) Why do we have to pass in an empty string for libraryIcon?  If it's optional, we should be able to pass in 
``` 
upLink: {
          url: new URL("http://up.com"),
          label: "Up Text",
          ariaLabel: "Up Aria Text",
        }
```

and just have it handle it.

hokei: D'oh.  This was what you had before. Sorry, I talked myself into this idea and now I've talked myself out of it.  

Originally I thought it would be good to go 
```
IFrameNavigator.create({
        element,
        manifestUrl: new URL("http://example.com/manifest.json"),
        store,
        settings,
        annotator,
        publisher,
        serif,
        sans,
        day,
        sepia,
        night,
        paginator,
        scroller,
        eventHandler,
      });
```

(notice: no upLink at all) and then it would fill in the defaults.  (The default URL would be to this git repo -- as it was in "vanilla" webpub-viewer).  The idea is, pretty much everybody would want to override this.  

But this makes it so that this example is less good, which hadn't occurred to me in the beginning so I think ... what you had before, is probably for the best.  

Sorry for asking you to run around in circles.

viewer/index.html
M0nica: Now it's setup so if a `libraryIcon`  is not passed in none of this information will display. I can update it to just send am empty string which will then display the default SVG. Before I was leaning towards libraryIcon being an optional value.

Another option is to add logic to treat SVGs differently from other types of files but adding SVG support would be more involved since the SVG would need to be edited to include attributes such as:
```
width = "${IconLib.WIDTH_ATTR}" height = "${IconLib.HEIGHT_ATTR}" viewBox = "${IconLib.VIEWBOX_ATTR} "aria-labelledby="up-label" 
```
I will check to see if we are expecting Open EBooks logo to be an SVG...

hokei: if SVG is complicated, let's do the simplest thing for now.  

I can imagine people wanting to do three things:  

1) I am a part of NYPL, I want to keep the NYPL icon but link to a different site (should use the SVG icon.  I think, best thing to do here is to accept 
```
upLink: {
          url: new URL("http://up.com"),
          label: "Up Text",
          ariaLabel: "Up Aria Text",
        }
```
and put the icon as the fallback svg
2) I am not a part of NYPL, I want to pass in my own logo and link.  I would pass in 
```
upLink: {
          url: new URL("http://up.com"),
          label: "Up Text",
          ariaLabel: "Up Aria Text",
          libraryIcon: new URL("icon.png") 
        }
```
3) I don't care, and don't want to configure this.  
Either, don't pass in upLink at all 
```
IFrameNavigator.create({
        element,
        manifestUrl: new URL("http://example.com/manifest.json"),
        store,
        settings,
        annotator,
        publisher,
        serif,
        sans,
        day,
        sepia,
        night,
        paginator,
        scroller,
        eventHandler,
      });
```
and it defaults to the library svg with placeholder link/label/arialabel,

or, still require something like this

```
upLink: {
          url: new URL("http://up.com"),
          label: "Up Text",
          ariaLabel: "Up Aria Text",
        }
```
to be passed in

Option 3 is the most nuanced to me, but also the option that matters the least - we should just pick something and go with it, I imagine most people will want to customize the logo, and those who don't will be ok just putting in placeholder values if they need to.

hokei (approved):

PR #167, opened 2020-07-24
Update to add .js extensions to imports

M0nica:
- .js imports still resolve to .ts in src
- explicitly make import extensions for dist version of webpub viewer .js


hokei:
I don't understand why this is necessary?  Shouldn't ts files be reading other ts files?

M0nica:
> I don't understand why this is necessary? Shouldn't ts files be reading other ts files?

I was running into issues when trying to use the `webpub-viewer`'s dist files in `circulation-patron-web` with the imports not being able to resolve. In `webpub-viewer`'s static example we were solving resolving the imports that by passing the express server  `extensions: ["js"]` so if an import was looking for `./Foo` and `./Foo` didn't exist it would also look for `./Foo.js` and resolve to that if it it exists.

M0nica:
I am going to close this as this was resolved by adding `const withTM = require("next-transpile-modules")([
  "library-simplified-webpub-viewer"
]);` to the next-config in the consumer (circulation-patron-web). 

- https://www.npmjs.com/package/next-transpile-modules 

- https://github.com/NYPL-Simplified/circulation-patron-web/blob/0c407352cf73e97822a479a8e1ede1cead44c227/next.config.js#L1

PR #168, opened 2020-08-21
Update page navigation buttons

M0nica:
- Show page navigation buttons when Table of Contents is inactive
- Hide page navigation buttons when Table of Contents is active (in order to resolve issue with page navigation buttons overlapping with the table of contents)

### DIFF
--- a/src/IFrameNavigator.ts
+++ b/src/IFrameNavigator.ts
@@ -1106,6 +1106,18 @@ export default class IFrameNavigator implements Navigator {
     }
   }
 
+  private hideNavigationButtons() {
+    const flipPageIcons = Array.from(
+      document.getElementsByClassName("flip-page-container")
+    );
+    for (const flipPageIcon of flipPageIcons) {
+      const newFlipPageClass =
+        (flipPageIcon.getAttribute("class") || "") + " hidden";
+
+      flipPageIcon.setAttribute("class", newFlipPageClass);
+    }
+  }
+
   private showModal(
     modal: HTMLDivElement,
     control?: HTMLAnchorElement | HTMLButtonElement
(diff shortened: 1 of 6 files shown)


src/IFrameNavigator.ts
```diff
@@ -1131,9 +1143,27 @@ export default class IFrameNavigator implements Navigator {
...
   }
 
+  private showNavigationButtons() {
+    const flipPageIcons = Array.from(
+      document.getElementsByClassName("flip-page-container")
+    );
+    for (const flipPageIcon of flipPageIcons) {
+      const newFlipPageClass = (
+        flipPageIcon.getAttribute("class") || ""
+      ).replace(" hidden", "");
```
EdwinGuzman, line 1161: I forgot how tedious it was to work directly with the DOM API.

EdwinGuzman:
The code looks good but I'm not sure how to manually test it. Just run the app, right?

M0nica:
> The code looks good but I'm not sure how to manually test it. Just run the app, right?

@EdwinGuzman Yep, you can manually test by:

1. running the `build` command and then the `static` command. 
2. Opening static app and click on Call of The Wild (Note: The app runs on `https`).
3. Test what happens to page navigation when the table of contents and open/closed and also test that other parts of the reader do not cause the page navigation to shift or change unexpectedly.

hokei:
Seems like there's been a regression where if you move to narrow screen, turn on the TOC, move to wide screen, turn off TOC, the reader is in a weird state.  I'm not able to reproduce this on SFR-develop

hokei:
but everything else looks good to me

src/IFrameNavigator.ts
hokei: yeah there's a reason why we just kept this in for months lol 😬

M0nica:
> but everything else looks good to me

@hokei I just updated the branch. I believe switching from `display: none` to `visibility:hidden` for `flip-page-container.hidden` resolves this issue.

hokei (approved):

PR #172, opened 2020-09-10
Resolve parsing errors for non-XHTML content

M0nica:
- Webpub viewer should only render XHTML content otherwise there are parsing errors and certain content doesn't render properly in the webpub viewer. This PR updates it so that the spine and TOC only contain content with the MediaType of XHTML and therefore only XHTML files are rendered within the viewer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,17 +4,28 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## [0.2.1]
+## [0.2.2]
+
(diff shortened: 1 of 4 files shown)


hokei:
does this work have anything to do with the fact that the chapter names appear badly, or is that another chunk of work that we should ticket?

hokei (approved):
I'm not seeing the chapter title errors on all books, so let's deal with this on a book-by-book basis during qa.

M0nica:
> does this work have anything to do with the fact that the chapter names appear badly, or is that another chunk of work that we should ticket?

I think that is a separate issue

PR #173, opened 2020-10-01
Fixes internal links

hokei:
Internal links used to throw a 404 error.  

This was because the webpub viewer assumed that the same server would serve both the parent and child of the iframe.  
Because this is no longer the case:  

1) Iframe srcdoc was modified to make sure that `base` referred to the site _inside_ the iframe
2) Same origin checking was changed to ensure referrer and link href were the same, rather than `window.location`.

### DIFF
--- a/src/EventHandler.ts
+++ b/src/EventHandler.ts
@@ -1,306 +1,346 @@
 import * as BrowserUtilities from "./BrowserUtilities";
 
 export default class EventHandler {
-    private pendingMouseEventStart: MouseEvent | null = null;
-    private pendingMouseEventEnd: MouseEvent | null = null;
-    private pendingTouchEventStart: TouchEvent | null = null;
-    private pendingTouchEventEnd: TouchEvent | null = null;
+  private pendingMouseEventStart: MouseEvent | null = null;
+  private pendingMouseEventEnd: MouseEvent | null = null;
+  private pendingTouchEventStart: TouchEvent | null = null;
+  private pendingTouchEventEnd: TouchEvent | null = null;
 
-    public onLeftTap: (event: UIEvent) => void = () => {};
-    public onMiddleTap: (event: UIEvent) => void = () => {};
-    public onRightTap: (event: UIEvent) => void = () => {};
-    public onBackwardSwipe: (event: UIEvent) => void = () => {};
-    public onForwardSwipe: (event: UIEvent) => void = () => {};
+  public onLeftTap: (event: UIEvent) => void = () => {};
+  public onMiddleTap: (event: UIEvent) => void = () => {};
+  public onRightTap: (event: UIEvent) => void = () => {};
+  public onBackwardSwipe: (event: UIEvent) => void = () => {};
+  public onForwardSwipe: (event: UIEvent) => void = () => {};
 
-    public onLeftArrow: (event: UIEvent) => void = () => {};
-    public onRightArrow: (event: UIEvent) => void = () => {};
+  public onLeftArrow: (event: UIEvent) => void = () => {};
+  public onRightArrow: (event: UIEvent) => void = () => {};
 
-    public onLeftHover: () => void = () => {};
-    public onRightHover: () => void = () => {};
-    public onRemoveHover: () => void = () => {};
+  public onLeftHover: () => void = () => {};
+  public onRightHover: () => void = () => {};
+  public onRemoveHover: () => void = () => {};
 
-    public onInternalLink: (event: UIEvent) => void = () => {};
+  public onInternalLink: (event: UIEvent) => void = () => {};
 
-    private static readonly CLICK_PIXEL_TOLERANCE = 10;
-    private static readonly TAP_PIXEL_TOLERANCE = 10;
-    private static readonly DOUBLE_CLICK_MS = 200;
-    private static readonly LONG_PRESS_MS = 500;
-    private static readonly DOUBLE_TAP_MS = 200;
-    private static readonly SLOW_SWIPE_MS = 500;
+  private static readonly CLICK_PIXEL_TOLERANCE = 10;
+  private static readonly TAP_PIXEL_TOLERANCE = 10;
+  private static readonly DOUBLE_CLICK_MS = 200;
+  private static readonly LONG_PRESS_MS = 500;
+  private static readonly DOUBLE_TAP_MS = 200;
+  private static readonly SLOW_SWIPE_MS = 500;
 
-    public setupEvents(element: HTMLElement | Document | null) {
-        if (element !== null) {
-            element.addEventListener("touchstart", this.handleTouchEventStart.bind(this));
-            element.addEventListener("touchend", this.handleTouchEventEnd.bind(this));
-            element.addEventListener("mousedown", this.handleMouseEventStart.bind(this));
-            element.addEventListener("mouseup", this.handleMouseEventEnd.bind(this));
-            // element.addEventListener("mouseenter", this.handleMouseMove.bind(this));
-            // element.addEventListener("mousemove", this.handleMouseMove.bind(this));
-            element.addEventListener("mouseleave", this.handleMouseLeave.bind(this));
+  public setupEvents(element: HTMLElement | Document | null) {
+    if (element !== null) {
+      element.addEventListener(
+        "touchstart",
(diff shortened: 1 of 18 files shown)


src/EventHandler.ts
```diff
@@ -269,11 +269,13 @@ export default class EventHandler {
...
         const link = this.checkForLink(event);
         if (link) {
+            const referrier = new URL(link.baseURI);
+
             // Open external links in new tabs.
             const isSameOrigin = (
-                window.location.protocol === link.protocol &&
-                window.location.port === link.port &&
-                window.location.hostname === link.hostname
+                referrier.protocol === link.protocol &&
```
kristojorg: Small typo

src/IFrameNavigator.ts
```diff
@@ -1368,27 +1368,22 @@ export default class IFrameNavigator implements Navigator {
...
-      this.iframe.contentDocument.location.href
-    ) {
-      currentLocation = this.iframe.contentDocument.location.href.split("#")[0];
-    }
-
     if (element && (element as HTMLElement).tagName.toLowerCase() === "a") {
-      if (
-        (element as HTMLAnchorElement).href.split("#")[0] === currentLocation
-      ) {
+        const href = (element as HTMLAnchorElement).href.split("#")[0];
```
kristojorg: I know this probably isn't part of this PR, and maybe isn't going to be changed ever if we eventually move to the readium web viewer, but in general these `as` statements should only be used when absolutely necessary, since they basically erase the safety benefits of using typescript. One way to make a value's type more specific without casting it, is to use a type guard, which is a function that checks whether it is of a certain type and narrows it to that type:
```ts
function isAnchorElement(element: HTMLElement): element is HTMLAnchorElement {
  return "tagName" in element && element.tagName.toLowerCase() === "a";
}
```
Something like that will then narrow the type for you from there on out

test/UtilsTest.ts
```diff
@@ -118,4 +118,18 @@ describe("Utils", () => {
             expect(embeddedXml).to.equal(xmlWithReplacedCSS)
         });
     })
+
+    describe.only("setBase", () => {
```
kristojorg: Should remove the `.only` here

kristojorg (changes requested):
I haven't tested this because I'm not totally sure how... maybe you can tell me? Do I need to npm link this with CPW and then view a read online book? But couple small things to fix, otherwise code looks good from my limited understanding

src/IFrameNavigator.ts
hokei: Honestly, not a bad idea to start putting these in and slowly moving them.  I have no idea when NYPL will move to the Readium web viewer, but I know there's still a lot of work planned for me to do on this.  "I'm gonna only be working on this for X more months so this QOL improvement isn't worth it" is a thing that has bitten me many times in the past.

hokei:
To Test:  `npm link library-simplified-webpub-viewer` in Circulation-Patron-Web to test links inside epubs.  Just a note:  If you update webpub-viewer and want to see the changes,  you have to run `npm run build` inside `webpub-viewer` to get it to show up.  
I need to find a webpub that has links inside it, so will test that on my end -- I want to make sure this hasn't broken that.  Ideally unit tests would catch that, but I'm not sure I fully trust the test coverage here.

src/IFrameNavigator.ts
kristojorg: Yea let's start adding them whenever we are updating code then

kristojorg:
Okay I'll give this a test asap then

EdwinGuzman:
Should this work with the open-ebooks url, `https://qa-circulation.openebooks.us/USOEI/groups/`? I've having a hard time testing because the "read online" link opens a blank page rather than the reader, and get this message:
```
There was a problem fetching this book. Please refresh the page or return home.

Error Code: unknown
Error Message: Failed to parse OPDS data
```

hokei:
How are you trying to test it?  The OPDS feed urls only work if you're running it inside circulation-patron-web.  on `npm run streamed` should still only work with webpub manifest urls (maybe direct links to epubs?  I havne't tried that in a while ...), and `npm run static` should only work with local books (epubs or webpubs)

EdwinGuzman:
Ah, do I need to start the server for the web reader? I was just running `cpw` but did `npm link` the package locally.

hokei:
You shouldn't need to - If you `npm link` the webreader you should be able to start cpw with the open ebooks params and see the local webreader.  If you're having issues msg me on slack and we can walk through it!

kristojorg:
When I link this and run it in CPW, and then navigate to this book: `https://localhost:3000/read/https%3A%2F%2Fqa-circulation.openebooks.us%2FUSOEI%2Fworks%2F114275%2Ffulfill%2F15`, and click a link to chapter 1, it triggers a download of a file it looks like. The link looks like this:
```
<a href="chapter001.xhtml"><strong>Begin Reading</strong></a>
```

And the filename for some reason looks like this:
```
![image](https://user-images.githubusercontent.com/4398775/95342247-864a8480-08b7-11eb-9aa9-932ae9300431.png)

```

kristojorg (changes requested):
When I link this and run it in CPW, and then navigate to this book: `https://localhost:3000/read/https%3A%2F%2Fqa-circulation.openebooks.us%2FUSOEI%2Fworks%2F114275%2Ffulfill%2F15`, and click a link to chapter 1, it triggers a download of a file it looks like. The link looks like this:
```
<a href="chapter001.xhtml"><strong>Begin Reading</strong></a>
```

And the filename for some reason looks like this:
```
![image](https://user-images.githubusercontent.com/4398775/95342247-864a8480-08b7-11eb-9aa9-932ae9300431.png)
```

Let me know if I'm testing the wrong thing though. The book is called "Rules of summer"

hokei:
oh no!  Let me take a look again, this is a pretty common issue that happens a lot when the reader state is weird.

hokei:
OK I neatened up the link event handler, I think this should work.

src/EventHandler.ts
```diff
@@ -1,304 +1,346 @@
...
+  //     } else {
+  //         this.onRemoveHover();
+  //     }
+  // }
+
+  private handleMouseLeave = (): void => {
+    this.onRemoveHover();
+  };
+
+  private handleLinks = (event: MouseEvent | TouchEvent): void => {
```
hokei, line 298: The new thing in this file is the addition of `isOnEpubHost` here.  I will tag the same thing in the original diff to help compare.  I'm also gonna file a "enforce a TS formatter" ticket because this is driving me nuts.

src/EventHandler.ts
```diff
@@ -1,304 +1,346 @@
...
-    // }
-
-    private handleMouseLeave = (): void => {
-        this.onRemoveHover();
+      // This was a single tap on a link. Do nothing.
+      this.pendingTouchEventEnd = null;
+      return;
     }
 
-    private handleLinks = (event: MouseEvent | TouchEvent): void => {
```
hokei, line 269: this was the method that changed

kristojorg:
I get this error when following an internal link on the book "Why Soccer Matters":

image

This is using the config branch in the PR I just opened. I will check if it also happens on beta... It looks like it is because there is a fetch request being sent to: `https://localhost:3000/openqa/read/9780698150058_EPUB-5.xhtml`... not sure if that's the problem or not.

kristojorg (approved):
Okay so it does work on beta with that book... I'll look in to why this isn't working on my latest PR. One thing is that on that same book, if you click "Dedication" in the table of contents, it downloads the dat file. I don't think that's blocking though. This seems okay to

kristojorg (changes requested):
Okay on second thought I think we need to figure out whats going on with the config branch before merging this. The HTML looks the same:
```
<a href="9780698150058_EPUB-5.xhtml">1</a>
```
But in the config branch it leads to a CPW link:
```
https://localhost:3000/openqa/read/9780698150058_EPUB-5.xhtml
```
 and in the beta branch it leads to:
```
https://node.axisnow.com/content/stream/9780698150058/OEBPS/9780698150058_EPUB-5.xhtml
```

So I don't really get thaaaaat

kristojorg (approved):
Okay after I cleared app storage in chrome devtools this now works on all branches! Woohoo!

PR #174, opened 2020-10-06
Errors don't show up by default

hokei:
The webpub viewer used to catch and display an error whenever one would show up, even when it got resolved later.  
It will now show the loading page 

When used in Circulation-patron-web, there are two loading states:  
cpw loads webpub-viewer 
webpub-viewer loads a book

This only addresses the second loading state.  It might be useful to put a circulation-patron-web loading screen when it is loading the viewer (in the future)

### DIFF
--- a/src/IFrameNavigator.ts
+++ b/src/IFrameNavigator.ts
@@ -20,6 +20,7 @@ import Encryption from "./Encryption";
 import Decryptor from "./Decryptor";
 import BookResourceStore from "./BookResourceStore";
 import { embedImageAssets, embedCssAssets } from "./Utils";
+
 const epubReadingSystemObject: EpubReadingSystemObject = {
   name: "Webpub viewer",
   version: "0.1.0",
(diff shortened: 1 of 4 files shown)


src/IFrameNavigator.ts
```diff
@@ -294,191 +295,185 @@ export default class IFrameNavigator implements Navigator {
 
   protected async start(element: HTMLElement, entryUrl: URL): Promise<void> {
     element.innerHTML = template;
-    try {
```
hokei, line 297: The main diff here is that I removed this try.  As a general rule I don't think that we should have such a big and generic try-catch.  I did keep the other ones in though because this app is built relying on them as a crutch and I'm not sure I understand potential errors well enough to feel good taking them out without going through all of the methods and doing error handling on them (which might take a lot longer time)

src/IFrameNavigator.ts
```diff
@@ -764,130 +759,121 @@ export default class IFrameNavigator implements Navigator {
   }
 
   private async loadManifest(): Promise<Manifest> {
-    try {
```
hokei, line 767: Same as the above comment.

hokei:
Moving this over from https://github.com/NYPL-Simplified/circulation-patron-web/pull/153

However it looks like the user isn't immediately recognized as having been logged in, so the login modal appears for about a second, and then it reloads. This only happens when you refresh the reader page or when you navigate to a reader page directly. It doesn't happen when you click "read online" from the catalog.

But again, if it's tricky let's prioritize getting this merged in and fix in a follow up.

src/IFrameNavigator.ts
```diff
@@ -941,89 +939,92 @@ export default class IFrameNavigator implements Navigator {
...
-          next.href,
-          this.manifestUrl.href
-        ).href;
-        this.nextChapterLink.className = "";
-      } else {
-        this.nextChapterLink.removeAttribute("href");
-        this.nextChapterLink.className = "disabled";
-        // this.handleRemoveHover();
-      }
+        const next = manifest.getNextSpineItem(currentLocation);
```
hokei, line 963: and this giant diff is because we don't have prettier installed.

kristojorg (approved):
Looks good to me, although the `package.json` is on a different version from `package-lock.json`. Also, would it make sense to add a CHANGELOG?

I think the best UX and DX for errors would be if the component is wrapped in an `ErrorBoundary` in CPW, which catches any errors thrown inside it and displays a fallback UI.

hokei:
> I think the best UX and DX for errors would be if the component is wrapped in an ErrorBoundary in CPW, which catches any errors thrown inside it and displays a fallback UI.

Ideally this package should stand alone, so it should handle its own errors ... it would be bad to expect that whoever's consuming it has to do something specific to make it work properly.  I'm curious though, why would this be better than figuring out what each method should do and what kinds of errors each should throw, and handling them on a case-by-case basis?  

Added changelog ... it looks like `package.json` and `package-lock.json` is the same ... but I'm not exactly sure we're looking at the same thing? Let me know if it still looks strange

PR #175, opened 2020-10-13
Chapter Title Parsing

hokei:
I filed https://jira.nypl.org/browse/SIMPLY-3186 as a follow up 
This only deals with the most surface of the issues, there are a bunch of underlying issues that may need fixing in the future.  The good news is that it shouldn't have broken anything already working.  The bad news is that I'm punting the writing of unit tests on this in favor of finishing up features by Wednesday.

### DIFF
--- a/src/IFrameNavigator.ts
+++ b/src/IFrameNavigator.ts
@@ -382,6 +382,52 @@ export default class IFrameNavigator implements Navigator {
     this.newElementId = null;
     this.isBeingStyled = true;
     this.isLoading = true;
+
+    var containerHref = entryUrl.href.endsWith("container.xml")
+      ? entryUrl.href
+      : "";
+    if (containerHref) {
+      this.manifestUrl = await Manifest.getManifestUrlFromContainer(
+        containerHref
+      );
+      //Check for existence of encryption doc.  A
+      // If a container is passed, assume epub.
+      const containerPath = entryUrl.href.substring(
+        0,
+        entryUrl.href.lastIndexOf("/")
+      );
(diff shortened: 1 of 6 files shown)


kristojorg:
I'm getting two errors on the config branch when trying to run this: `cannot read href of undefined` and `cannot read length of undefined`. I tried clearing my browser data to no avail :/

kristojorg:
Also wtf IFrameNavigator.ts is 1700 lines?! I wish I didn't see that

src/IFrameNavigator.ts
```diff
@@ -876,16 +804,115 @@ export default class IFrameNavigator implements Navigator {
...
+              position: 0,
+            });
+          }
+        }
+      });
+
+      parentElement.appendChild(listElement);
+    };
+
+    if (manifest.toc.length) {
```
kristojorg, line 877: This might be source of the `cannot read length of undefined` error. Just a guess

kristojorg (changes requested):
Reading through this code I am sorry for what you're having to do :(

hokei:
> Also wtf IFrameNavigator.ts is 1700 lines?! I wish I didn't see that

:trash_fire_emoji:

kristojorg (approved):
Woohoo we are in business! It looks good to me. Some books don't have the TOC in the header, but presumably that's because they're not compatible

hokei:
> Some books don't have the TOC in the header, but presumably that's because they're not compatible

Yeah in my initial survey I thought most of the books were epub3 but it turns out a bunch of them aren't.  I'll add them in next week.

PR #176, opened 2020-10-14
Added support for an additional font

hokei:
This is just using a placeholder from google fonts -- will replace with the URL of OpenDyslexia when we get that up.

### DIFF
--- a/src/DyslexiaFont.ts
+++ b/src/DyslexiaFont.ts
@@ -11,7 +11,16 @@ export default class DyslexiaFont implements BookFont {
     const rootFrame = HTMLUtilities.findRequiredIframeElement(this.bookElement.contentDocument, "html") as HTMLHtmlElement;
 
     HTMLUtilities.setAttr(rootFrame, "data-viewer-font", "dyslexia-friendly");
-    HTMLUtilities.createStylesheet(rootFrame, "dyslexia-font-internal", "* {font-family: 'Dancing Script', cursive !important;}");
+    HTMLUtilities.createStylesheet(rootFrame, "dyslexia-font-stylesheet", 
+    `@font-face {
+      font-family: 'opendyslexic';
+      src: url('https://nypl-static.s3.amazonaws.com/base/fonts/OpenDyslexic3-Regular.woff2') format('woff2'),
+           url('https://nypl-static.s3.amazonaws.com/base/fonts/OpenDyslexic3-Regular.woff') format('woff');
+      font-weight: normal;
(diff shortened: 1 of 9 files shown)


src/IFrameNavigator.ts
```diff
@@ -234,6 +237,7 @@ export default class IFrameNavigator implements Navigator {
     (document as any).msFullscreenEnabled;
 
   public static async create(config: IFrameNavigatorConfig) {
+    console.log("config", config);
```
kristojorg: Extra console

test/IFrameNavigatorTests.ts
```diff
@@ -2594,7 +2594,7 @@ describe("IFrameNavigator", () => {
       const titlePage = `<?xml version="1.0" encoding="UTF-8" ?>
 <html>
 <head>
-<title>Book Tittle</title>
```
kristojorg, line 2606: I prefer tittle

kristojorg (commented):
Is this meant to be enabled in the settings panel? I'm not seeing it there...

kristojorg (approved):
Nevermind, I ran it with the right CPW branch and it worked!

hokei:
> Is this meant to be enabled in the settings panel? I'm not seeing it there...


Yeah, you gotta enable it when you call the reader, and I think it doesn't need to be enabled by default on webpub-viewer, but it /should/ live in webpub viewer since it's something that non-CPW people might want to offer!

src/DyslexiaFont.ts
```diff
@@ -0,0 +1,32 @@
...
+  public readonly label = "Dyslexia Friendly";
+
+  public bookElement: HTMLIFrameElement;
+
+  public start(): void {
+    const rootFrame = HTMLUtilities.findRequiredIframeElement(this.bookElement.contentDocument, "html") as HTMLHtmlElement;
+
+    HTMLUtilities.setAttr(rootFrame, "data-viewer-font", "dyslexia-friendly");
+    HTMLUtilities.createStylesheet(rootFrame, "dyslexia-font-stylesheet", 
+    `@font-face {
```
hokei, line 15: yes, this is the way to load a new font inside this iframe.   No, I haven't figured out how to do this via a css file.

kristojorg:
Appears to be working on local with the newest beta

kristojorg (approved):

src/DyslexiaFont.ts
kristojorg: lol

PR #177, opened 2020-10-21
Added prettier for better linting

hokei:
At long last. Most of these options I copied over from `circulation-patron-web`


.eslintignore
```diff
@@ -0,0 +1,5 @@
+viewer
+dist
+lib
+
+webpub-rnow/
```
kristojorg, line 5: Are these included in the gitignore as well? If so I don't think you need them here just fyi. Not a big deal tho

kristojorg (approved):
Looks good to me, I didn't look through every change but I looked at the config files and then assume you just ran prettier to generate all the others, in which case this should be good to go!

kristojorg:
Although I wonder why travis is failing?

hokei:
I have no idea why it failed the last time, I hate it. 

It's weird, not putting it in eslintignore was making it so that it was trying to lint those files.  Aaah, I'll just keep them they're not hurting anyone haha

PR #178, opened 2020-10-22
Epub 2 TOC

hokei:
Adds support for table of contents when provided only in Epub2 format.  

An example of a book that works in this scheme but not before is "Starfall"


kristojorg (approved):
Wow magical . The reader is like, looking pretty useable now!

PR #179, opened 2020-10-26
Pre-loading screen and async book chapter loading

hokei:
this PR does two things:  
1) Fetches all the book resources at once, rather than fetching them one at a time
2) Puts the loading spinner while resources are fetching.  (The delay is an old artifact, do you think it is a good idea to show the spinner right away?) 

I tried it with a few books, but please try it with a few more to help me verify that this all actually works!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## UNRELEASED
+## [0.2.5]
 ### Adds 
 - Prettier to format code
+
+### Fixes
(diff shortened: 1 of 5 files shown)


hokei:
navigating between chapters with arrows seems to not be working, will look into this in a bit

kristojorg (approved):
This is really really good. Speeds up load times for me by a lot! I think the next step is a loading indicator in CPW on the `/read` page (this will get rid of the blank screen before the axisnow loader shows up I think).

I don't fully understand the code obviously, but I didn't see anything glaring. 

The one book that didn't work for me is "Exile for Dreamers", but I just checked and it also doesn't work on beta

kristojorg:
Does this need a changelog though?

hokei:
Oh of course!

PR #180, opened 2020-11-09
Fixed persistent arrow and extra margins

hokei:


kristojorg (approved):
Looks good to me! I haven't seen the exact problem this is trying to solve but this seems to look find on the books I checked with

